PR #17, opened 2023-05-08
add custom RateLimitError message for auth users

mrteey:


flavioschneider:
Hi! I didn't set a default error for this exception because it's from the API error message, see: https://github.com/elevenlabs/elevenlabs-python/blob/4fa487be76b0e4b57c6b233af63867c129df3f3b/elevenlabs/api/base.py#L49 

This will show an error message which also contains information about the character limit.

mrteey:
> Hi! I didn't set a default error for this exception because it's forwarded from the API error message, see:
> 
> https://github.com/elevenlabs/elevenlabs-python/blob/4fa487be76b0e4b57c6b233af63867c129df3f3b/elevenlabs/api/base.py#L49
> 
> This will show an error message which also contains information about the character limit.

Oh I get it, but then why weren't you just using the APIError class since it already handles that instead of creating another class that simply inherits and adds nothing?

flavioschneider:
RateLimitError is raised only when you reach the rate limit. If we raised the more generic APIError then the user would not be able to catch the specific case :) (e.g. if you have a UI and want to personalise what happens when rate limit is reached)

PR #41, opened 2023-06-03
New functions for unsupported endpoints + some modifications

Sxvxgee:
This PR brings support to multiple endpoints that I noticed were not supported in this wrapper:
1. `/v1/voices/settings/default`
2. `/v1/voices/{voice_id}/settings`
3. `/v1/voices/{voice_id}/settings/edit`
4. `/v1/voices/{voice_id}`
5. `/v1/voices/{voice_id}/edit`
6. `/v1/history/{history_item_id}`
7. `/v1/user/subscription`

I've also modified the following:
1. `Subscription` class: Added new properties.
2. `User` class: the `from_api` function was not working so I fixed it.

### DIFF
--- a/elevenlabs/api/history.py
+++ b/elevenlabs/api/history.py
@@ -57,6 +57,9 @@ class HistoryItem(API):
         if self._audio is None:
             self._audio = API.get(url).content
         return self._audio
+    
+    def delete(self):
+        API.delete(f"{api_base_url_v1}/history/{self.history_item_id}")
 
 
 class History(Listable, API):
--- a/elevenlabs/api/user.py
+++ b/elevenlabs/api/user.py
@@ -1,11 +1,21 @@
 from __future__ import annotations
+from typing import Optional
 
 from .base import API, api_base_url_v1
 
-
 class Subscription(API):
+    tier: str
     character_count: int
     character_limit: int
+    can_extend_character_limit: bool
+    allowed_to_extend_character_limit: bool
+    next_character_count_reset_unix: int
+    voice_limit: int
+    professional_voice_limit: int
+    can_extend_voice_limit: bool
+    can_use_instant_voice_cloning: bool
+    can_use_professional_voice_cloning: bool
+    currency: Optional[str]
     status: str
 
 
@@ -16,4 +26,9 @@ class User(API):
     def from_api(cls) -> User:
         url = f"{api_base_url_v1}/user"
         response = cls.get(url).json()
-        return cls(**response)
+        return cls(subscription=Subscription(**response.get("subscription", {})))
+    
+    def get_subscription(cls):
+        url = f"{api_base_url_v1}/user/subscription"
+        response = API.get(url).json()
+        return Subscription(**response)
--- a/elevenlabs/api/user.py
+++ b/elevenlabs/api/user.py
@@ -18,6 +18,12 @@ class Subscription(API):
     currency: Optional[str]
     status: str
 
+    @classmethod
+    def from_api(cls) -> Subscription:
+        url = f"{api_base_url_v1}/user/subscription"
+        response = API.get(url).json()
+        return cls(**response)
+
 
 class User(API):
     subscription: Subscription
@@ -26,9 +32,4 @@ class User(API):
     def from_api(cls) -> User:
         url = f"{api_base_url_v1}/user"
         response = cls.get(url).json()
-        return cls(subscription=Subscription(**response.get("subscription", {})))
-    
-    def get_subscription(cls):
-        url = f"{api_base_url_v1}/user/subscription"
-        response = API.get(url).json()
-        return Subscription(**response)
+        return cls(**response)
--- a/elevenlabs/api/voice.py
+++ b/elevenlabs/api/voice.py
@@ -133,36 +133,35 @@ class Voice(API):
     def computed_settings(cls, v: VoiceSettings, values) -> VoiceSettings:
         url = f"{api_base_url_v1}/voices/{values['voice_id']}/settings"
         return v if v else VoiceSettings(**API.get(url).json())
+    
+    @classmethod
+    def default_settings(cls):
+        url = f"{api_base_url_v1}/voices/settings/default"
+        return VoiceSettings(**API.get(url).json())
 
     def delete(self):
         API.delete(f"{api_base_url_v1}/voices/{self.voice_id}")
 
-    def edit_settings(self, stability: float, similarity_boost: float):
+    # def edit_settings(self, stability: float, similarity_boost: float):
+    def edit_settings(self, voice_settings: VoiceSettings):
         url = f"{api_base_url_v1}/voices/{self.voice_id}/settings/edit"
         json = {
-            "stability": stability,
-            "similarity_boost": similarity_boost,
+            "stability": voice_settings.stability,
+            "similarity_boost": voice_settings.similarity_boost,
         }
         API.post(url, json=json)
-
-    def get_settings(self):
-        url = f"{api_base_url_v1}/voices/{self.voice_id}/settings"
-        return VoiceSettings(**API.get(url).json())
     
-    def get_info(self):
-        url = f"{api_base_url_v1}/voices/{self.voice_id}"
-        return Voice(**API.get(url).json())
-    
-    def edit(self, name: str, labels: Optional[str] = None, description: Optional[str] = None):
-        url = f"{api_base_url_v1}/voices/{self.voice_id}/edit"
-        data = {
-            "name": name,
-        }
-        if labels: # Putting it as None will overwrite the labels
-            data["labels"] = labels
-        if description: # Putting it as None will overwrite the description
-            data["description"] = description
-        API.post(url, data=data)
+    def edit(
+        cls,
+        name: Optional[str] = None,
+        labels: Optional[str] = None,
+        description: Optional[str] = None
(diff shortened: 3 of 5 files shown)


elevenlabs/api/voice.py
```diff
@@ -137,6 +137,32 @@ def computed_settings(cls, v: VoiceSettings, values) -> VoiceSettings:
...
+            "stability": stability,
+            "similarity_boost": similarity_boost,
+        }
+        API.post(url, json=json)
+
+    def get_settings(self):
+        url = f"{api_base_url_v1}/voices/{self.voice_id}/settings"
+        return VoiceSettings(**API.get(url).json())
+    
+    def get_info(self):
```
flavioschneider, line 152: Same as above

elevenlabs/simple.py
```diff
@@ -81,6 +81,10 @@ def voices() -> Voices:
     VOICES_CACHE = Voices.from_api(api_key) if api_key else VOICES_CACHE
     return VOICES_CACHE
 
+def default_settings() -> VoiceSettings:
```
flavioschneider, line 84: I woudn't add this less common use in the simple api

elevenlabs/api/voice.py
```diff
@@ -146,6 +172,11 @@ def from_api(cls, api_key: Optional[str] = None):
         url = f"{api_base_url_v1}/voices"
         response = API.get(url).json()
         return cls(**response)
+    
+    @classmethod
+    def default_settings(cls):
```
flavioschneider, line 177: Rethinking about this maybe it's better to put it into the `Voice` object? As `Voice.default_settings()`

elevenlabs/api/user.py
```diff
@@ -1,11 +1,21 @@
 from __future__ import annotations
+from typing import Optional
 
 from .base import API, api_base_url_v1
 
-
 class Subscription(API):
+    tier: str
```
flavioschneider, line 7: This is good:)

elevenlabs/api/history.py
```diff
@@ -57,6 +57,9 @@ def audio(self) -> bytes:
         if self._audio is None:
             self._audio = API.get(url).content
         return self._audio
+    
+    def delete(self):
```
flavioschneider, line 61: Good!

elevenlabs/api/voice.py
```diff
@@ -137,6 +137,32 @@ def computed_settings(cls, v: VoiceSettings, values) -> VoiceSettings:
...
+
+    def get_settings(self):
+        url = f"{api_base_url_v1}/voices/{self.voice_id}/settings"
+        return VoiceSettings(**API.get(url).json())
+    
+    def get_info(self):
+        url = f"{api_base_url_v1}/voices/{self.voice_id}"
+        return Voice(**API.get(url).json())
+    
+    def edit(self, name: str, labels: Optional[str] = None, description: Optional[str] = None):
```
flavioschneider, line 156: This can get out of sync with the `self` object, e.g. name can be different from the one in the API. I'd do something like 
```py 
self.name = name or self.name,
self.labels = labels or self.labels,
self.description = description or self.description

data = dict(name=self.name, labels=self.labels, description=self.description) 
```

where also the `name` input is optional. This also allows the user to manually change the voice object, e.g. `voice.name='new name',` and then call `voice.edit()`

flavioschneider:
Thanks for the PR! I added some comments for things I'd do differently

Sxvxgee:
> Thanks for the PR! I added some comments for things I'd do differently

You're welcome! Also, I can't see any comment. Have you submitted the review?

elevenlabs/api/user.py
```diff
@@ -16,4 +26,9 @@ class User(API):
     def from_api(cls) -> User:
         url = f"{api_base_url_v1}/user"
         response = cls.get(url).json()
-        return cls(**response)
+        return cls(subscription=Subscription(**response.get("subscription", {})))
```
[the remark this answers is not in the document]
Sxvxgee: Ah interesting. So should I undo it or just keep it as is?

elevenlabs/api/user.py
```diff
@@ -16,4 +26,9 @@ class User(API):
...
+    
+    def get_subscription(cls):
```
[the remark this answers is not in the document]
Sxvxgee: Alright, I'll do that!

elevenlabs/api/voice.py
```diff
@@ -137,6 +137,32 @@ def computed_settings(cls, v: VoiceSettings, values) -> VoiceSettings:
     def delete(self):
         API.delete(f"{api_base_url_v1}/voices/{self.voice_id}")
 
+    def edit_settings(self, stability: float, similarity_boost: float):
```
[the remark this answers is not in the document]
Sxvxgee: So you mean making a parameter in the function with the `VoiceSettings` object as the type?

elevenlabs/api/voice.py
```diff
@@ -137,6 +137,32 @@ def computed_settings(cls, v: VoiceSettings, values) -> VoiceSettings:
...
+        url = f"{api_base_url_v1}/voices/{self.voice_id}/settings/edit"
+        json = {
+            "stability": stability,
+            "similarity_boost": similarity_boost,
+        }
+        API.post(url, json=json)
+
+    def get_settings(self):
```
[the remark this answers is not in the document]
Sxvxgee: That is true, I haven't thought of it. I'll remove it as well as `get_info`.

elevenlabs/api/user.py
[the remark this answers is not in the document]
flavioschneider: I'd undo this part but keep the new fields in Subscription object

elevenlabs/api/voice.py
Sxvxgee: Thanks for bringing that to my attention, I agree with you and I'll implement these changes. However, pardon my question, but did you mean that you want me to use `cls` instead of `self`? Also, I was wondering, could you tell me what the purpose of `@validator("settings")` is? I never saw such a thing  before.

elevenlabs/api/voice.py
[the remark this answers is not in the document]
flavioschneider: Yeah exactly: `edit_settings(voice_settings: VoiceSettings)`

elevenlabs/api/voice.py
Sxvxgee: Fair enough, will do that soon.

elevenlabs/api/user.py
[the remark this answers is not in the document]
Sxvxgee: Alright thank you for your response, I'll do that!

elevenlabs/api/voice.py
flavioschneider: Depends on whether you want to modify the current object or any voice object using it's id, if you want you can do both `edit(self, name, labels, ...)` to edit the instance object and `edit(cls, voice_id, name,...)` which can be called as `Voice.edit(voice_id, name, ...)`. 

The `@validator("settings")` basically calls the `computed_settings` function when the voice object is created, this is to do some custom initialization of the `settings` field. The function provides as input the current value of the `settings: VoiceSettings` when the class is initiazlied, then you can do whatever you want to it and return the updated value. In this case, since the API doesn't return the settings of the voice when doing `Voice.from_id(..)`, we use this to automatically fetch the /settings endpoint, and populate the `settings` field. If instead you manually instantiate a voice object with `Voice(settings=VoiceSettings(...))` the input to that function will not be `None` and hence nothing will happen: `v if v else VoiceSettings(**API.get(url).json())` (see the `if` check)

Sxvxgee:
I just pushed 3 commits that do as you've requested :)

Sxvxgee:
Also may I ask why you use `dict()` instead of `{}`?

flavioschneider:
Thanks! I like `dict()` more because the keys are not written as strings

elevenlabs/api/voice.py
```diff
@@ -133,10 +133,35 @@ def from_design(cls, voice_design: VoiceDesign):
...
+        url = f"{api_base_url_v1}/voices/settings/default"
+        return VoiceSettings(**API.get(url).json())
 
     def delete(self):
         API.delete(f"{api_base_url_v1}/voices/{self.voice_id}")
 
+    # def edit_settings(self, stability: float, similarity_boost: float):
+    def edit_settings(self, voice_settings: VoiceSettings):
+        url = f"{api_base_url_v1}/voices/{self.voice_id}/settings/edit"
+        json = {
```
flavioschneider, line 148: here you can do: `json = voice_settings.dict()`

elevenlabs/api/voice.py
```diff
@@ -133,10 +133,35 @@ def from_design(cls, voice_design: VoiceDesign):
...
+            "stability": voice_settings.stability,
+            "similarity_boost": voice_settings.similarity_boost,
+        }
+        API.post(url, json=json)
+    
+    def edit(
+        cls,
```
flavioschneider, line 155: If you use `cls` you must provide the id as input `def edit(cls, voice_id,...)` if you want to use the id form the self object then you have to use `self` instead

flavioschneider (commented):
Added a few more comments

Sxvxgee:
Done!

flavioschneider:
Perfect, thanks for the PR!

PR #42, opened 2023-06-05
Add Usage on Root README for Advanced Usage with Voice Settings

isurewould:
# What
Adding an "Advanced Usage" example on the README Page

# Why
When using the "Speech Synthesis" UI tab it is quite easy to change the Voice Settings. For users of the python lib an example using Voice Settings should be prominent on the main README

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,7 +104,7 @@ from elevenlabs.api.voice import Voice, VoiceSettings
 
 # define custom voice settings for Bella
 voice = Voice(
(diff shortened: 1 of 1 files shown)


isurewould:
Up to you guys. 🤷  

Just a suggestion. ✌️ 

-Ian

flavioschneider:
I would add this example in the `API.md` file. I might edit the simple `generate` function to include voice settings to make it simpler for users.

isurewould:
@flavioschneider sure I can move API.md

PR #60, opened 2023-07-19
Update error.py, use first index of list as detail

### LINKED ISSUE #54: TypeError in error.py
In the HTTPError class init function, I am getting:
`TypeError: list indices must be integers or slices, not str`
This happens because the first line of the function is: `detail = json.loads(response.text)["detail"]` 
This is returning a list and NOT a dict. The dict is the first index of the list.

Additionally, in the same function, it tries to set the variables self.message and self.status to detail["message"] and detail["status"}, but those keys do NOT exist. The key that I think self.message needs is actually "msg", and there is NO key for the self.status variable.

 Also, in the final line of the init function, `self.additional_info = detail.get("additional_info", None)`, is being called on the list, not the dict it is expecting, which is the first index of the list.

felivalencia3:
Fixes #54

### DIFF
--- a/elevenlabs/api/error.py
+++ b/elevenlabs/api/error.py
@@ -10,10 +10,11 @@ class HTTPError:
     additional_info: Optional[Dict] = None
 
     def __init__(self, response: requests.Response):
-        detail = json.loads(response.text)["detail"]
-        self.message = detail["message"]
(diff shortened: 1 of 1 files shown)


flavioschneider:
Thanks!

PR #63, opened 2023-07-21
added latency param

fylls:
optimize_streaming_latency is missing in the generate_stream function. read more here: https://help.elevenlabs.io/hc/en-us/articles/15726761640721-Can-I-reduce-API-latency-


fylls:
my bad here https://github.com/elevenlabs/elevenlabs-python/pull/64 you can find the real pull request.

the #63 did not work locally, sorry for uploading untested code lmao

PR #74, opened 2023-08-05
HTTPError doesn't have a .message property

monological:


flavioschneider:
This has been fixed in `v0.2.25`.

Issue #67, opened 2023-07-29
async support?


bryanhpchiang:
for real-time applications / things with an ongoing event loop.

the TTS generate + streaming calls are both IO bound so it seems like this would make sense

marty-sullivan:
Yes, would be great to use this official library, but we cannot use this on a web server because it has no async methods.

amitkot:
@flavioschneider does providing asyncio python support currently on your roadmap?
We would appreciate un update 🙏

dsinghvi:
@bryanhpchiang ElevenLabs just published a beta v3 SDK that you can download [here](https://github.com/elevenlabs/elevenlabs-python?tab=readme-ov-file#%EF%B8%8F-install). The SDK now exports a sync and async client!

PR #80, opened 2023-08-10
Fully address issue #67 - Async API Support

gvariable:
# Description:
This pull request addresses Issue #67 by introducing async API support to the project. The changes implemented in this PR enable concurrent operations and enhance the overall efficiency of the system.

# Key Changes
- Added async support based on [httpx](https://www.python-httpx.org/), which offers an API similar to the widely used requests library.
- Implemented async support for user, voice, model, tts, and history components.
- Developed test cases to validate the functionality of the newly added async API.
- Ensured a seamless transition between sync and async code with a similar API structure like `User.afrom_api()`.

In the future, there is a possibility of incorporating support for async websockets to enable streaming requests. This enhancement would allow for more efficient and dynamic communication.

### DIFF
--- a/elevenlabs/simple.py
+++ b/elevenlabs/simple.py
@@ -1,6 +1,6 @@
 import os
 import re
-from typing import Iterator, Optional, Union
+from typing import Iterator, AsyncIterator, Optional, Union
 
 from .api import TTS, Model, Voice, VoiceClone, Voices
 from .cache import VOICES_CACHE
@@ -39,7 +39,6 @@ def generate(
     latency: int = 1,
     stream_chunk_size: int = 2048,
 ) -> Union[bytes, Iterator[bytes]]:
-
     if isinstance(voice, str):
         voice_str = voice
         # If voice is valid voice_id, use it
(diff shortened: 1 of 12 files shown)


elevenlabs/simple.py
```diff
@@ -72,3 +71,46 @@ def generate(
...
+
+async def agenerate(
+    text: Union[str, Iterator[str]],
+    api_key: Optional[str] = None,
+    voice: Union[str, Voice] = VOICES_CACHE[2],  # Bella
+    model: Union[str, Model] = "eleven_monolingual_v1",
+    stream: bool = False,
+    latency: int = 1,
+    stream_chunk_size: int = 2048,
+) -> Union[bytes, AsyncIterator[bytes]]:
```
wheresthecode, line 84: I'd suggest moving the majority of this code into a utility function since it is duplicated between the generate function

wheresthecode:
I'd like to see this feature added to the API as well. Just browsing through the code it looks like there is a lot of duplicated logic now. If the existing code were refactored a bit, perhaps the code could be unified. This applies to both tests and code.

elevenlabs/simple.py
```diff
@@ -72,3 +71,46 @@ def generate(
     else:
         assert isinstance(text, str)
         return TTS.generate(text, voice, model, api_key=api_key)
+
+
+async def agenerate(
+    text: Union[str, Iterator[str]],
```
filip-michalsky, line 77: Shouldn't this be also `AsyncIterator`?

Rickaym:
Awesome work. I feel like it would also be helpful to include an async AsyncIterator input to AsyncInterator output. Sort of like 
```
def generate(
    text: str | Iterator[str],
    api_key: str | None = None,
    voice: str | Voice = VOICES_CACHE[2],
    model: str | Model = "eleven_monolingual_v1",
    stream: bool = False,
    latency: int = 1,
    stream_chunk_size: int = 2048
) -> (bytes | Iterator[bytes])
```

How the generate() method can take Iterator and returns an Iterator, aka, taking a stream of text and streaming an audio in return.

bitnom:
Is any help needed here? This is an obvious thing to want to merge.

shachar-ug:
+1

shachar-ug:
C'mon guys, merge :+1:

praetor29:
I played around with this branch in a fork, and it isn't _quite_ easy as hitting merge, but I think if someone qualified takes a look at the code, it's 90% there.

marty-sullivan:
Yes, would be great to use this official library, but we cannot use this on a web server because it has no async methods.

frmsaul:
Thanks for working on this!

dsinghvi:
Hey all -- ElevenLabs now uses [Fern](https://buildwithfern.com/) to automatically generate SDKs using their OpenAPI spec. As part of this, the latest SDK exports both a sync and async client! Please give it a shot using the installation instructions in the [README](https://github.com/elevenlabs/elevenlabs-python?tab=readme-ov-file#%EF%B8%8F-install).

PR #86, opened 2023-08-16
Added description and labels fields to VoiceDesign

### LINKED ISSUE #54: TypeError in error.py

### LINKED ISSUE #94: from_design() method not working with sample code
Using the [sample code](https://github.com/elevenlabs/elevenlabs-python/blob/main/API.md#voicedesign) in the docs, the [`from_design()` method](https://github.com/elevenlabs/elevenlabs-python/blob/3d1f80a3f33a76c9b84054313df5e3e5f5a96726/elevenlabs/api/voice.py#L118) is returning the following error. It looks like the method's API call is responding with a list whereas a dictionary or JSON object is expected.

```
Traceback (most recent call last):
  File "/Users/daniel/Documents/python/elevenlabs/save-voice.py", line 25, in <module>
    voice = elevenlabs.Voice.from_design(design)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/daniel/Library/Python/3.11/lib/python/site-packages/elevenlabs/api/voice.py", line 128, in from_design
    response = API.post(url, json=data)
               ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/daniel/Library/Python/3.11/lib/python/site-packages/elevenlabs/api/base.py", line 65, in post
    return API.request(url, method="post", *args, **kwargs)  # type: ignore
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/daniel/Library/Python/3.11/lib/python/site-packages/elevenlabs/api/base.py", line 46, in request
    error = HTTPError(response)
            ^^^^^^^^^^^^^^^^^^^
  File "/Users/daniel/Library/Python/3.11/lib/python/site-packages/elevenlabs/api/error.py", line 14, in __init__
    self.message = detail["message"]
                   ~~~~~~^^^^^^^^^^^
TypeError: list indices must be integers or slices, not str
```

ephraimacquah:
The server-side requirements for `v1/voice-generation/create-voice` include a description and labels field. The mismatch between the previous setup of `VoiceDesign` and the server-side requirements was causing HTTPError422, Unprocessable Entity.
Updated API.md to match the new design of `VoiceDesign`
closes #54
closes #94

### DIFF
--- a/elevenlabs/api/voice.py
+++ b/elevenlabs/api/voice.py
@@ -68,6 +68,8 @@ class VoiceDesign(API):
     name: str
     text: str = Field(..., min_length=100)
     gender: Gender
+    description: str
+    labels: dict
     age: Age
     accent: Accent
     accent_strength: float = Field(..., gt=0.3, lt=2.0)
@@ -123,7 +125,9 @@ class Voice(API):
         url = f"{api_base_url_v1}/voice-generation/create-voice"
         data = dict(
             voice_name=voice_design.name,
+            voice_description=voice_design.description,
             generated_voice_id=voice_design.generated_voice_id,
+            labels=voice_design.labels,
         )
         response = API.post(url, json=data)
         voice = cls.from_id(voice_id=response.json()["voice_id"])
--- a/API.md
+++ b/API.md
@@ -272,7 +272,7 @@ from elevenlabs import Voice, VoiceDesign, Gender, Age, Accent, play
 design = VoiceDesign(
     name='Lexa',
     text="Hello, my name is Lexa. I'm your personal assistant, I can help you with your daily tasks and I can also read you the news.",
-    description="A young british female voice, resembling a secretary. Perfect for the office assistant roles.",
+    description="A young british female voice, resembling a secretary. Perfect for office assistant roles.",
     labels={'accent': 'british', 'gender': 'female', 'age': 'young'},
     gender=Gender.female,
     age=Age.young,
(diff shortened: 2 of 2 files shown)


lpabon (approved):
This is great! We use this library in our project https://github.com/aipresenter/aipresenter and found this issue. We were able to debug it using the Javascript sources of the Web UI and found the solution there.

elevenlabs/api/voice.py
```diff
@@ -68,6 +68,8 @@ class VoiceDesign(API):
     name: str
     text: str = Field(..., min_length=100)
     gender: Gender
+    description: str
+    labels: dict
```
lpabon, line 72: Does `labels` take any value as keys or is it looking for specific keys?

ephraimacquah: labels use specific keys being age, gender, and accent

flavioschneider:
Hi, thanks! Voice design has been fixed in `v0.2.25`.

PR #120, opened 2023-10-05
add new feature tts stream async context manager

linebe-eu:


pencil:
@dunky11 @flavioschneider Any chance to get this looked at? Streaming data is the perfect use case for async Python so having official support for it would be great.

PR #143, opened 2023-11-22
Add output_format to client library.

MarkoMackic:
One question remains, as per API docs, mp3 output format for websocket is only mp3_44100 , but does it also support specifying bit rate ?


efiShtain:
Would be really helpful to see this change on main

PR #217, opened 2024-03-21
solving asynchronization issue in AsyncTextToSpeechClient in both convert and convert_as_stream

Mahmoud-ghareeb:
Both **convert** and **convert_as_stream** functions had an issue while using the "AsyncElevenlabs" class. The issue appeared as follows:

when I tried to do the following:

`
client = AsyncElevenLabs(api_key=...)
audio_stream = await client.generate(text="how are you", stream=True, ....)
async for i in audio_stream:
    print(I)
`
I got this error:

**coroutine object doesn't support the asynchronous context manager protocol**

The issue was solved by adding **await** within the "convert and convert_as stream" functions in the "AsyncTextToSpeechClient" class.

### DIFF
--- a/src/elevenlabs/text_to_speech/client.py
+++ b/src/elevenlabs/text_to_speech/client.py
@@ -114,7 +114,8 @@ class TextToSpeechClient:
                         "optimize_streaming_latency": optimize_streaming_latency,
                         "output_format": output_format,
                         **(
-                            request_options.get("additional_query_parameters", {})
+                            request_options.get(
+                                "additional_query_parameters", {})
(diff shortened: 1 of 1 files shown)


Mahmoud-ghareeb:
Ok @fern-bot thank you.

Mahmoud-ghareeb:
@fern-bot  I have a Code Formatter extension in my VS code that's why all these changes happened.
the added code in lines 383 and 519

Mahmoud-ghareeb:
I wanted to be added as contributor 😫

PR #374, opened 2024-10-15
feat: add issue templates

louisjoecodes:
Adds structured templates to submit either a feedback request/bug fix, new issues are now automatically tagged on creation.

TODO: Make sure people do not use issues for security incident reports.

### DIFF
--- a/.github/ISSUE_TEMPLATE/1-bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/1-bug_report.yml
@@ -1,0 +1,51 @@
+name: Bug report
+description: Create a bug report for the ElevenLabs Python SDK
+labels: [bug]
+body:
+  - type: markdown
+    attributes:
+      value: |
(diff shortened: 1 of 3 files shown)


lacop11 (approved):
Thanks

lacop11:
> TODO: Make sure people do not use issues for security incident reports.

Yeah we should point people to the right place for security issues, good call. I'll figure out the link (might have to make one) & ping you after, can be included in the template.

.github/ISSUE_TEMPLATE/1-bug_report.yml
```diff
@@ -0,0 +1,51 @@
...
+
+        **Expected behavior:**
+
+        **Actual behavior:**
+    validations:
+      required: true
+  - type: textarea
+    attributes:
+      label: Code example
+      description: Provide an example code snippet that has the problem
```
lacop11, line 35: Maybe add a warning to remove/censor the API key?

PR #379, opened 2024-10-17
chore/update bb policy

louisjoecodes:
- chore minor security policy updates


lacop11 (approved):
Thanks, LGTM. Let's follow up on adding the contact email as soon as it is ready, it's a bit silly to say "let us know right away" but then not provide a way to do so :D

louisjoecodes:
Absolutely!

PR #389, opened 2024-10-24
Implement basic SDK for convai agents

lacop11:
Early prototype, subject to change based on user feedback.

Takes care of the websocket session and message handling, exposing a simplified audio interface to the client that can be hooked up to the appropriate audio inputs / outputs based on the usecase.

Also implements a basic speaker/microphone interface, via optional dependency on pyaudio.

### DIFF
--- a/src/elevenlabs/conversation.py
+++ b/src/elevenlabs/conversation.py
@@ -1,0 +1,289 @@
+from abc import ABC, abstractmethod
+import base64
+import json
+import queue
+import threading
+from typing import Callable, Optional
+
+from websockets.sync.client import connect
+
+from .base_client import BaseElevenLabs
+
+
+class AudioInterface(ABC):
+    """AudioInterface provides an abstraction for handling audio input and output."""
+
+    @abstractmethod
+    def start(self, input_callback: Callable[[bytes], None]):
+        """Starts the audio interface.
+
+        Called one time before the conversation starts.
+        The `input_callback` should be called regularly with input audio chunks from
+        the user. The audio should be in 16-bit PCM mono format at 16kHz. Recommended
+        chunk size is 4000 samples (250 milliseconds).
+        """
+        pass
+
+    @abstractmethod
+    def stop(self):
+        """Stops the audio interface.
+
+        Called one time after the conversation ends. Should clean up any resources
+        used by the audio interface and stop any audio streams. Do not call the
+        `input_callback` from `start` after this method is called.
+        """
+        pass
+
+    @abstractmethod
+    def output(self, audio: bytes):
+        """Output audio to the user.
+
+        The `audio` input is in 16-bit PCM mono format at 16kHz. Implementations can
+        choose to do additional buffering. This method should return quickly and not
+        block the calling thread.
+        """
+        pass
+
+    @abstractmethod
+    def interrupt(self):
+        """Interruption signal to stop any audio output.
+
+        User has interrupted the agent and all previosly buffered audio output should
+        be stopped.
+        """
+        pass
+
+
+class Conversation:
+    client: BaseElevenLabs
+    agent_id: str
+    requires_auth: bool
+
+    audio_interface: AudioInterface
+    callback_agent_response: Optional[Callable[[str], None]]
+    callback_user_transcript: Optional[Callable[[str], None]]
+    callback_latency_measurement: Optional[Callable[[int], None]]
+
+    _thread: Optional[threading.Thread] = None
+    _should_stop: threading.Event = threading.Event()
+    _conversation_id: Optional[str] = None
+    _last_interrupt_id: int = 0
+
+    def __init__(
+        self,
+        client: BaseElevenLabs,
+        agent_id: str,
+        *,
+        requires_auth: bool,
+        audio_interface: AudioInterface,
+        callback_agent_response: Optional[Callable[[str], None]] = None,
+        callback_agent_response_correction: Optional[Callable[[str, str], None]] = None,
+        callback_user_transcript: Optional[Callable[[str], None]] = None,
+        callback_latency_measurement: Optional[Callable[[int], None]] = None,
+    ):
+        """Conversational AI session.
+
+        BETA: This API is subject to change without regard to backwards compatibility.
+
+        Args:
+            client: The ElevenLabs client to use for the conversation.
+            agent_id: The ID of the agent to converse with.
+            requires_auth: Whether the agent requires authentication.
+            audio_interface: The audio interface to use for input and output.
+            callback_agent_response: Callback for agent responses.
+            callback_agent_response_correction: Callback for agent response corrections.
+                First argument is the original response (previously given to
+                callback_agent_response), second argument is the corrected response.
+            callback_user_transcript: Callback for user transcripts.
+            callback_latency_measurement: Callback for latency measurements (in milliseconds).
+        """
+
+        self.client = client
+        self.agent_id = agent_id
+        self.requires_auth = requires_auth
+
+        self.audio_interface = audio_interface
+        self.callback_agent_response = callback_agent_response
+        self.callback_agent_response_correction = callback_agent_response_correction
+        self.callback_user_transcript = callback_user_transcript
+        self.callback_latency_measurement = callback_latency_measurement
+
+    def start_session(self):
+        """Starts the conversation session.
+
+        Will run in background thread until `end_session` is called.
+        """
+        ws_url = self._get_signed_url() if self.requires_auth else self._get_wss_url()
+        self._thread = threading.Thread(target=self._run, args=(ws_url,))
+        self._thread.start()
+
+    def end_session(self):
+        """Ends the conversation session."""
+        self.audio_interface.stop()
+        self._should_stop.set()
+
+    def wait_for_session_end(self) -> Optional[str]:
+        """Waits for the conversation session to end.
+
+        You must call `end_session` before calling this method, otherwise it will block.
+
+        Returns the conversation ID, if available.
+        """
+        if not self._thread:
+            raise RuntimeError("Session not started.")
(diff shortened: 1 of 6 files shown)


src/elevenlabs/conversation.py
```diff
@@ -0,0 +1,282 @@
...
+        pass
+
+
+class Conversation:
+    client: BaseElevenLabs
+    agent_id: str
+    requires_auth: bool
+
+    audio_interface: AudioInterface
+    callback_agent_response: typing.Optional[typing.Callable[[str], None]]
```
jodik, line 63: I would import these from typing to increase readability

jodik, line 63: we should add callback_agent_response_correction

src/elevenlabs/conversation.py
```diff
@@ -0,0 +1,282 @@
...
+        """Interruption signal to stop any audio output.
+
+        User has interrupted the agent and all previosly buffered audio output should
+        be stopped.
+        """
+        pass
+
+
+class Conversation:
+    client: BaseElevenLabs
```
jodik, line 58: This will work if empty ElevenLabs() is passed without api key correct?

src/elevenlabs/conversation.py
```diff
@@ -0,0 +1,282 @@
...
+        """
+
+        self.client = client
+        self.agent_id = agent_id
+        self.requires_auth = requires_auth
+
+        self.audio_interface = audio_interface
+        self.callback_agent_response = callback_agent_response
+        self.callback_user_transcript = callback_user_transcript
+        self.callback_latency_measurement = callback_latency_measurement
```
jodik, line 104: Since most are passed by name, what about pydantic BaseModel? Or you think too restrictive?

src/elevenlabs/conversation.py
```diff
@@ -0,0 +1,282 @@
...
+                            "user_audio_chunk": base64.b64encode(audio).decode(),
+                        }
+                    )
+                )
+
+            self.audio_interface.start(input_callback)
+            while not self._should_stop.is_set():
+                try:
+                    message = json.loads(ws.recv(timeout=0.5))
+                    self._handle_message(message, ws)
```
jodik, line 148: check for is_set here will prohibit message parsing past `stop()`

src/elevenlabs/conversation.py
```diff
@@ -0,0 +1,282 @@
...
+
+def get_default_audio_interface():
+    """Returns a default audio interface implementation.
+
+    Uses the default system input and output sound devices.
+    """
+    try:
+        import pyaudio
+    except ImportError:
+        raise ImportError("To use the default audio interface, you must install and pyaudio.")
```
jodik, line 213: This could be part of __init__() and users could just directly import Default(Audio)Interface instead of the get_default_audio_interface, no?

src/elevenlabs/conversation.py
jodik, line 1: I would follow the rest of the directory and put this under `conversational_ai/` subfolder, then we can move `DefaultAudioInterface` (which most enterprises won't use) into utils/default_audio_interface.py and keep focused core at `conversation.py`

Note in the future more files corresponding to various endpoints e.g. `agents.py` will be added.

jodik (approved):
Great! thanks a lot Laco

src/elevenlabs/conversation.py
lacop11: Yes, as long as auth is not required. We still need the client to construct the server endpoint URL (and ElevenLabs() allows overriding that).

src/elevenlabs/conversation.py
lacop11: I considered it but we don't have pydantic as dependency now and so I don't want to add it just to save myself few lines. Any dependency we take is additional constraint on the users of this package.

src/elevenlabs/conversation.py
```diff
@@ -0,0 +1,282 @@
...
+                stream_callback=self._in_callback,
+                frames_per_buffer=self.INPUT_FRAMES_PER_BUFFER,
+                start=True,
+            )
+            self.out_stream = self.p.open(
+                format=pyaudio.paInt16,
+                channels=1,
+                rate=16000,
+                output=True,
+                frames_per_buffer=self.OUTPUT_FRAMES_PER_BUFFER,
```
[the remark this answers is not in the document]
lacop11: The docs aren't super clear on this (they clearly chose the golang way of comments 
image
)
but basically smaller buffer allows us to cut the agent faster - the stuff already in the audio buffer we sent to pyaudio can't be changed so it will play. 

I'm not entirely satisfied with it yet, could eg. do a smoother volume crossfade instead of immediate interruption. But this change should not impact the public API so I'll save that for later, what we have is good enough for first demos.

src/elevenlabs/conversation.py
```diff
@@ -0,0 +1,282 @@
...
+            self.p.terminate()
+
+        def output(self, audio: bytes):
+            self.output_queue.put(audio)
+
+        def interrupt(self):
+            # Clear the output queue to stop any audio that is currently playing.
+            try:
+                while True:
+                    _ = self.output_queue.get(block=False)
```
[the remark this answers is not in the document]
lacop11: Nope. Importantly this blocks the message parsing thread so there is no race between clearing & new audio being added. I've added a comment.

src/elevenlabs/conversation.py
lacop11: Well that is the idea, `end_session` will close the websocket and the audio streams. Do you want some grace period here or what is the concern?

src/elevenlabs/conversation.py
lacop11: Missed that one, added now.

src/elevenlabs/conversation.py
lacop11: Sure, updated.

src/elevenlabs/conversation.py
```diff
@@ -0,0 +1,282 @@
...
+        ws_url = self._get_signed_url() if self.requires_auth else self._get_wss_url()
+        self._thread = threading.Thread(target=self._run, args=(ws_url,))
+        self._thread.start()
+
+    def end_session(self):
+        """Ends the conversation session."""
+        self.audio_interface.stop()
+        self._should_stop.set()
+
+    def wait_for_session_end(self) -> typing.Optional[str]:
```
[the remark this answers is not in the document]
lacop11: It's not really strictly needed but if combined one method would request the end and also block on it which takes a while. This way there is a blocking wait and a non-blocking method to request end that returns immediately so can be used as signal handler etc

src/elevenlabs/conversation.py
lacop11: Hmm yes but I need to confirm what the autogenerated code will look like, I don't want to conflict with that (if I put this in location that fern will want to autogenerate it will be a problem and we don't want to keep manually updating it)

lacop11:
Thanks for reviews, addressed most of the feedback.

src/elevenlabs/conversation.py
lacop11: OK I've checked, there will be `src/elevenlabs/conversational_ai/client.py` which will have all the current openapi generated calls. So it is safe to put this in `conversational_ai/conversation.py`. I've also split of the `default_audio_interface.py` but I kept it in the same folder rather than `utils/`.

lacop11:
Moved the code & updated demo code https://gist.github.com/lacop11/c29cc81fea5fe02b3e291bc12b9e9a31

src/elevenlabs/conversation.py
jodik: I would add the check right before handle message again. 

Time 0: while not is_set() -> True
Time 0.1: stop()
Time 0.5: ws.recv() returns message
Time 0.501:  message is still handled

src/elevenlabs/conversational_ai/conversation.py
```diff
@@ -0,0 +1,210 @@
...
+
+class Conversation:
+    client: BaseElevenLabs
+    agent_id: str
+    requires_auth: bool
+
+    audio_interface: AudioInterface
+    callback_agent_response: Optional[Callable[[str], None]]
+    callback_user_transcript: Optional[Callable[[str], None]]
+    callback_latency_measurement: Optional[Callable[[int], None]]
```
jodik, line 64: callback_agent_response_correction missing here?

src/elevenlabs/conversational_ai/conversation.py
```diff
@@ -0,0 +1,210 @@
...
+            audio = base64.b64decode(event["audio_base_64"])
+            self.audio_interface.output(audio)
+        elif message["type"] == "agent_response":
+            if self.callback_agent_response:
+                event = message["agent_response_event"]
+                self.callback_agent_response(event["agent_response"].strip())
+        elif message["type"] == "agent_response_correction":
+            if self.callback_agent_response_correction:
+                event = message["agent_response_correction_event"]
+                self.callback_agent_response_correction(event["original_agent_response"].strip(), event["corrected_agent_response"].strip())
```
jodik, line 174: Is this indented correctly? Github shows weirdly

jodik (approved):
Good to go

src/elevenlabs/conversational_ai/conversation.py
lacop11: Oops, thanks for catching

src/elevenlabs/conversational_ai/conversation.py
lacop11: It's fine, just long line. Ran `ruff format` which split it up a bit.

src/elevenlabs/conversation.py
lacop11: Ah I see you want the opposite. OK added even though it probably doesn't matter that much.

PR #410, opened 2024-12-03
fix: stream with timestamp now returns a value

### LINKED ISSUE #396: TextToSpeechClient.stream_with_timestamps does not return or yield response
### Description

I was trying to use Python SDK to have a stream with timestamps. 
But turns out that `TextToSpeechClient.stream_with_timestamps` does not yield chunks of HTTP response and returns `None` instead of `typing.Iterator[bytes]`.

Here is the place where I suppose something should be yielded or returned:
https://github.com/elevenlabs/elevenlabs-python/blob/main/src/elevenlabs/text_to_speech/client.py#L637

For example, neighbor method `convert_as_stream` yields chunks of specified size:
https://github.com/elevenlabs/elevenlabs-python/blob/main/src/elevenlabs/text_to_speech/client.py#L488

Unless I am not able to see some meta magic, same piece of code as in `convert_as_stream` should do the job. But I do not know, how can I send a PR to autogenerated code or do something further.

### Code example

```
client = ElevenLabs(api_key=settings.ELEVENLABS_API_KEY)

audio_stream = client.text_to_speech.stream_with_timestamps(
    text=text,
    voice_id=settings.ELEVENLABS_VOICE_ID,
    model_id='eleven_multilingual_v2',
)

# audio_stream is None, but I expected typing.Iterator[bytes] or something like that.
```

### Additional context

If someone has the same problem while this is being fixed, you can subclass original `TextToSpeechClient`, override `stream_with_timestamps` with suggested fix and then use it directly, e.g. like:

```
client = ElevenLabs(api_key=settings.ELEVENLABS_API_KEY)
tts = TextToSpeechClient(client_wrapper=client.text_to_speech._client_wrapper)

audio_stream = tts.stream_with_timestamps(
    text=text,
    voice_id=settings.ELEVENLABS_VOICE_ID,
    model_id='eleven_multilingual_v2',
)
```

rootux:
WIP - ⚠️ don't merge

# What
Fixed #396 - TextToSpeechClient.stream_with_timestamps does not return or yield response

# Why
I believe it was simply forgotten in the original implementation

# 🎩 Tophatting
Tested that indeed when calling this function - we are receiving back response (See screenshot)

# Design Alternatives
This function is called `stream` - but there isn't any stream - it receive a single text - perhaps think about renaming it / allowing both `text` / `stream`.

p.s I've noticed that your JS library has the same bug of not returning any value from that function.

# Screenshots

Screenshot 2024-12-03 at 8 59 36

### DIFF
--- a/src/elevenlabs/text_to_speech/client.py
+++ b/src/elevenlabs/text_to_speech/client.py
@@ -538,7 +538,7 @@ class TextToSpeechClient:
             BodyTextToSpeechStreamingWithTimestampsV1TextToSpeechVoiceIdStreamWithTimestampsPostApplyTextNormalization
         ] = OMIT,
         request_options: typing.Optional[RequestOptions] = None,
-    ) -> None:
+    ) -> typing.Optional[typing.Any]:
         """
         Converts text into speech using a voice of your choice and returns a stream of JSONs containing audio as a base64 encoded string together with information on when which character was spoken.
 
@@ -647,7 +647,13 @@ class TextToSpeechClient:
         )
         try:
             if 200 <= _response.status_code < 300:
-                return
+                return typing.cast(
+                    typing.Optional[typing.Any],
+                    construct_type(
+                        type_=typing.Optional[typing.Any],  # type: ignore
+                        object_=_response.json(),
+                    ),
+                )
             if _response.status_code == 422:
                 raise UnprocessableEntityError(
                     typing.cast(


louisjoecodes:
Checking

louisjoecodes:
Hello @rootux could you please check-out this (https://github.com/elevenlabs/elevenlabs-python/pull/412) branch and let me know if this fixes the issue? This should negate the need to merge this PR and close this: https://github.com/elevenlabs/elevenlabs-python/issues/396

rootux:
Perfect.
Thanks. I just noticed that indeed this is a stream.
Yes that pr you shared indeed solves my issue

louisjoecodes:
Have you tried it? Just making sure the types are correct before committing :)

rootux:
@louisjoecodes yep it's working :)
Just noticed you changed somewhere it was `audio_base64` and now it's `audio_base_64`. but either way it was broken before, so I think it's working good now ✅

rootux:
@louisjoecodes when tested - only the first chunk in the stream returns the full object with alignments. 
Also - the audio there doesn't contain the full alignments.

Rest of the chunks - alignments are None. only the audio exists.
Not sure if intended or a bug.

PR #411, opened 2024-12-03
Fix/stream subtitles

rootux:


louisjoecodes:
Hello! What was the goal here? Note that our Python SDK is fully auto-generated from an OpenAPI specification. Is there some missing functionality

PR #418, opened 2024-12-06
fix: move class-level attributes to instance-level in Conversation class

### LINKED ISSUE #417: Multiple conversations interfere due to class-level attributes
### What I was trying to achieve
---
I am running a FastAPI server which initializes a new `Conversation` instance for each incoming call connected via Twilio’s WebSocket streaming. I expected each `Conversation` instance to operate entirely independently, allowing multiple concurrent conversations without interference.

### What happened instead
---
After one conversation ended, its `_should_stop` event remained set. Because `_should_stop` is defined as a class-level attribute, this event was effectively shared by all `Conversation` instances. As a result, any new or ongoing conversations started after the first one ended behaved as if they were also stopped, preventing further audio processing or communication.

### Steps to reproduce
---
1. Start a Conversation instance (Conversation A) and begin a session.
2. End Conversation A (via `end_session()`)
3. Start a second Conversation instance (Conversation B)

### Bug source
---
- File: `elevenlabs/conversational_ai/conversation.py`

### Suggested fix
---
Define the following attributes as instance-level (inside the __init__ method) rather than at the class level:

- `_thread`
- `_should_stop`
- `_conversation_id`
- `_last_interrupt_id`

### Code example

```python
import os
import signal

from elevenlabs.client import ElevenLabs
from elevenlabs.conversational_ai.conversation import Conversation
from elevenlabs.conversational_ai.default_audio_interface import DefaultAudioInterface

api_key = <>
agent_id = <>

client = ElevenLabs(api_key=api_key)

conversation_1 = Conversation(
    client,
    agent_id,
    requires_auth=False,
    audio_interface=DefaultAudioInterface(),
)

print(f'\n_should_stop conversation_1 before start: {conversation_1._should_stop.is_set()}\n')

conversation_1.start_session()
signal.signal(signal.SIGINT, lambda sig, frame: conversation_1.end_session())
conversation_1.wait_for_session_end()

conversation_2 = Conversation(
    client,
    agent_id,
    requires_auth=False,
    audio_interface=DefaultAudioInterface(),
)

print(f'\n_should_stop conversation_2 before start: {conversation_2._should_stop.is_set()}\n')

conversation_2.start_session()
signal.signal(signal.SIGINT, lambda sig, frame: conversation_2.end_session())
conversation_2.wait_for_session_end()
```

### Additional context

_No response_

acse-efk23:
Fixes #417 

Why:
Class-level attributes persisted state between sessions. Making them instance-level ensures each session starts fresh.

What:
Moved `_thread`, `_should_stop`, `_conversation_id`, and `_last_interrupt_id` from class-level to instance-level within the __init__ method.

Note: This pull request is a proof of concept, demonstrating how moving class-level attributes to instance-level resolves the session interference issue. I understand this can’t be merged directly due to the code generation process.

### DIFF
--- a/src/elevenlabs/conversational_ai/conversation.py
+++ b/src/elevenlabs/conversational_ai/conversation.py
@@ -73,10 +73,10 @@ class Conversation:
     callback_user_transcript: Optional[Callable[[str], None]]
     callback_latency_measurement: Optional[Callable[[int], None]]
 
-    _thread: Optional[threading.Thread] = None
+    _thread: Optional[threading.Thread]
     _should_stop: threading.Event
-    _conversation_id: Optional[str] = None
-    _last_interrupt_id: int = 0
+    _conversation_id: Optional[str]
+    _last_interrupt_id: int
 
     def __init__(
         self,
@@ -119,7 +119,11 @@ class Conversation:
         self.callback_agent_response_correction = callback_agent_response_correction
         self.callback_user_transcript = callback_user_transcript
         self.callback_latency_measurement = callback_latency_measurement
+
+        self._thread = None
         self._should_stop = threading.Event()
+        self._conversation_id = None
+        self._last_interrupt_id = 0
 
     def start_session(self):
         """Starts the conversation session.


louisjoecodes:
Thank you @acse-efk23! @tudor11l,  @lacop11 - do you think this is something we want to merge? Makes sense to isolate these at instance-level rather than class-level

tudor11l:
@louisjoecodes we fixed the only one that was causing a problem which was the mutable (by not initialising it with a default value)
```threading.Event = threading.Event()```

I'm not a huge fan of class-level variables especially when they do not have the same value across instances. But it depends on what others think as well.

lacop11:
Thanks @acse-efk23 !

We have fixed the mutable `_should_stop` issue (I didn't notice your PR earlier unfortunately, would save us some time debugging!) but I agree we need a more consistent approach.

Note that these are not class variables but instance variables, IIUC. See https://docs.python.org/3/library/typing.html#typing.ClassVar.

The problem was the mutable default value. I think we should be fine just dropping _all_ default values  and only keeping type hints on the class level, then putting all defaults into `__init__`. We can do it for the `= None` and `= 0` even though they are not mutable and don't cause issues, just to be consistent and avoid accidentally adding more mutable defaults in future.

@acse-efk23 BTW: 
> I understand this can’t be merged directly due to the code generation process.
This particular file is not generated and is excluded in https://github.com/elevenlabs/elevenlabs-python/blob/main/.fernignore#L4 so it won't be overwritten.

If you would like to adjust the PR to keep the type hints as they were and just move the value assignment we would be happy to merge it. :)

acse-efk23:
Thank you for the review, @lacop11 . I have updated the pull request accordingly.

lacop11:
Looks good but the branch is out of date and conflicts, can you please rebase + resolve?

acse-efk23:
Thank you, @lacop11 . I have rebased and solved the conflicts.

lacop11 (approved):
Thanks, looks good!

PR #426, opened 2024-12-09
Move _should_stop initialization to __init__

tudor11l:
Avoid using mutable default values for _should_stop by initializing it in the __init__ method to ensure instance-level independence.


lacop11 (approved):
Thanks!

PR #427, opened 2024-12-09
Removes the hardcoded websocket api path.

Nickronomicon:


src/elevenlabs/realtime_tts.py
```diff
@@ -39,6 +40,10 @@ def text_chunker(chunks: typing.Iterator[str]) -> typing.Iterator[str]:
 
 
 class RealtimeTextToSpeechClient(TextToSpeechClient):
+    def __init__(self, *, client_wrapper: SyncClientWrapper):
+        super().__init__(client_wrapper=client_wrapper)
+        parsed = urllib.parse.urlparse(self._client_wrapper.get_base_url())
+        self._ws_base_url = urllib.parse.ParseResult("wss", *parsed[1:]).geturl()
```
lacop11, line 46: Nit: maybe `base_url.replace(scheme="wss")` ?

lacop11 (approved):
I'm assuming you tested this :)

src/elevenlabs/realtime_tts.py
Nickronomicon: on a named tuple, the method is `_replace` and I wasn't sure how happy we are with calling "private" methods, especially on a public repo. If we're OK with it, I can make that change.

Nickronomicon:
> I'm assuming you tested this :)

😁  Tested with the explicit base_urls https://api.elevenlabs.io and https://nick-upgrade-fastapi---xi-backend-server-dev-6ual7pllha-uc.a.run.app
Also tested the default with no suppled base_url, they all work.

src/elevenlabs/realtime_tts.py
lacop11: https://docs.python.org/3/library/urllib.parse.html#:~:text=As%20is%20the%20case%20with%20all%20named%20tuples%2C%20the%20subclass%20has%20a%20few%20additional%20methods%20and%20attributes%20that%20are%20particularly%20useful.%20One%20such%20method%20is%20_replace().%20The%20_replace()%20method%20will%20return%20a%20new%20ParseResult%20object%20replacing%20specified%20fields%20with%20new%20values.

Docs call this out and gives example so I think it's fine to use.

Nickronomicon: oh, cool and good. I like this better.

PR #445, opened 2024-12-17
feat: Client Tools

louisjoecodes:
Add support for client-tools in Python

### DIFF
--- a/src/elevenlabs/conversational_ai/conversation.py
+++ b/src/elevenlabs/conversational_ai/conversation.py
@@ -2,7 +2,9 @@ from abc import ABC, abstractmethod
 import base64
 import json
 import threading
-from typing import Callable, Optional
+from typing import Callable, Optional, Awaitable, Union, Any
+import asyncio
+from concurrent.futures import ThreadPoolExecutor
 
 from websockets.sync.client import connect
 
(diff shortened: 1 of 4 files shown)


src/elevenlabs/conversational_ai/conversation.py
```diff
@@ -52,22 +52,44 @@ def interrupt(self):
         """
         pass
 
+
+class ClientTools:
+    def __init__(self):
+        self.tools = {}
+        self.lock = threading.Lock()
+
+    def register(self, tool_name, handler):
```
lacop11, line 61: Can you add type annotation here please?

Also given the call blocks the processing of audio maybe add a comment saying the tool call should finish quickly. I think eventually we should turn this interface async and support async tools, then we can return the response when ready.

tests/e2e_test_convai.py
```diff
@@ -0,0 +1,71 @@
...
+
+    # Get API key from environment
+    api_key = os.getenv("ELEVENLABS_API_KEY")
+    if not api_key:
+        raise ValueError("Please set ELEVENLABS_API_KEY environment variable")
+
+    # Initialize client
+    client = ElevenLabs(api_key=api_key)
+
+    AGENT_ID = "<insert-testing-agent-id>"
```
lacop11, line 22: Take from env var maybe?

lacop11 (approved):
Nice, thanks for adding the tests!

src/elevenlabs/conversational_ai/conversation.py
louisjoecodes: Good shout! added.

PR #448, opened 2024-12-22
pyproject.toml: Add Homepage and Issues URLs, bump Python versions to currently supported ones

C0rn3j:
Drops 3.8 and sets 3.9 as minimum, explicitly adds support for 3.13 as according to the current upstream Support status - https://devguide.python.org/versions/

Adds Homepage and Issues links so that one can click-through them from PyPI website.

Fixes Repository to point to a .git repo as per https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#a-full-example

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -30,7 +30,7 @@ packages = [
 
 [project.urls]
 Homepage = "https://github.com/elevenlabs/elevenlabs-python"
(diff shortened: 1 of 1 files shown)


C0rn3j:
I don't see relevant changes in the main branch, I presume it's present somewhere else?

Issue #457, opened 2025-01-22
Manage end_call tool in Conversational AI / Twilio integration


futilm:
### Description

Hello,

I encountered an exception while using the provided code for Conversational AI / Twilio integration available [here](https://elevenlabs.io/docs/conversational-ai/guides/conversational-ai-twilio#implementation). In particular, the exception occurs when you enable the end_call tool for the agent used in the call. 

Exception:

```
Error in input_callback: received 1000 (OK); then sent 1000 (OK)
Exception in thread Thread-1 (_run):
Traceback (most recent call last):
  File "/usr/lib/python3.10/threading.py", line 1016, in _bootstrap_inner
Error in input_callback: received 1000 (OK); then sent 1000 (OK)
Error in input_callback: received 1000 (OK); then sent 1000 (OK)
Error in input_callback: received 1000 (OK); then sent 1000 (OK)
Error in input_callback: received 1000 (OK); then sent 1000 (OK)
Error in input_callback: received 1000 (OK); then sent 1000 (OK)
    self.run()
  File "/usr/lib/python3.10/threading.py", line 953, in run
    self._target(*self._args, **self._kwargs)
  File "/home/[REDACTED]/.local/share/virtualenvs/[REDACTED]/lib/python3.10/site-packages/elevenlabs/conversational_ai/conversation.py", line 295, in _run
    message = json.loads(ws.recv(timeout=0.5))
  File "/home/[REDACTED]/.local/share/virtualenvs/[REDACTED]/lib/python3.10/site-packages/websockets/sync/connection.py", line 291, in recv
    raise self.protocol.close_exc from self.recv_exc
websockets.exceptions.ConnectionClosedOK: received 1000 (OK); then sent 1000 (OK)
Error in input_callback: received 1000 (OK); then sent 1000 (OK)
Error in input_callback: received 1000 (OK); then sent 1000 (OK)
# [...] Multiple lines of this until CTRL+C
Ending conversation session...
```

It appears the provided code is unaware of the possible websocket `ConnectionClosedOK` exception.

How should I modify the code in order to handle a connection close event FROM the agent during a Twilio call?

louisjoecodes:
Hey thanks for flagging, checking this.

louisjoecodes:
This should be resolved!

futilm:
Hi @louisjoecodes ,

thank you, I confirm it works.

[...]

PR #462, opened 2025-01-29
fix: handle end call gracefully

AngeloGiacco:
currently, if we get a websocket connection closed OK exception, the client blows up.
instead, it now calls end_session to ensure that client tools/audio get cleaned up

fixes https://github.com/elevenlabs/elevenlabs-python/issues/457


louisjoecodes (approved):
Great! I think the WebSocket could also raise other exceptions ConnectionClosedError, WebSocketException, OSError, etc. so we may want to handle those in another PR eventually.

Issue #434, opened 2024-12-13
WebSocket connection terminates unexpectedly during conversation in Python SDK demo


jcassady:
## Description
When running the ElevenLabs Python SDK Conversational AI demo on a Raspberry Pi 5, the WebSocket connection terminates unexpectedly after a few conversation exchanges. The agent successfully responds to the first two conversation exchanges, but during the third exchange, the voice agent's reply is cut off mid-response, followed by a WebSocket connection closure error.

## Steps to reproduce
1. Set up environment variables (`AGENT_ID` and `ELEVENLABS_API_KEY`)
2. Run the demo script from elevenlabs-examples repository
3. Begin conversation with the agent
4. After approximately 2-3 exchanges, the connection fails

## What happened instead
The WebSocket connection closes unexpectedly with a timeout error (0.5 seconds) in the websocket receive operation, terminating the conversation prematurely:

```bash
Exception in thread Thread-1 (_run):
...
  File ".../elevenlabs/conversational_ai/conversation.py", line 152, in _run
    message = json.loads(ws.recv(timeout=0.5))
                         ^^^^^^^^^^^^^^^^^^^^
...
```
See **additional context** for the complete console log.

## Relevant parameters or configurations
- Running on Raspberry Pi 5
- OS: Debian GNU/Linux 12 (bookworm)
- Kernel: 6.6.62+rpt-rpi-2712
- Architecture: aarch64
- Audio devices:
  - USB2.0 Device (card 2)  # speaker
  - CMTECK (card 3) # mic

### poetry run python convai/demo.py 
```bash
joelcassady@jcpi:~/git/elevenlabs-examples/examples/conversational-ai/python $ poetry run python convai/demo.py 
ALSA lib pcm_asym.c:105:(_snd_pcm_asym_open) capture slave is not defined
ALSA lib confmisc.c:1369:(snd_func_refer) Unable to find definition 'cards.0.pcm.front.0:CARD=0'
ALSA lib conf.c:5180:(_snd_config_evaluate) function snd_func_refer returned error: No such file or directory
ALSA lib conf.c:5703:(snd_config_expand) Evaluate error: No such file or directory
ALSA lib pcm.c:2666:(snd_pcm_open_noupdate) Unknown PCM front
ALSA lib pcm.c:2666:(snd_pcm_open_noupdate) Unknown PCM cards.pcm.rear
[...]

PR #463, opened 2025-01-30
fix: add general error handling

AngeloGiacco:


AngeloGiacco:
handles ConnectionClosedError and other errors like referenced in #434

louisjoecodes (approved):

Issue #464, opened 2025-01-30
System type in Prompt Agent Tools not working in Create Agent API


salman-aziz-4425:
### Description

<img width="548" alt="Image" src="https://github.com/user-attachments/assets/39c82acf-0b84-4f16-b1d8-95849a5babcf" />

<img width="934" alt="Image" src="https://github.com/user-attachments/assets/81b3f6ba-7597-4d2a-854c-0cd211a34e56" />

it do not show 'system' type and throws error to enforce client type usage

### Code example

    agent_config = AgentConfig(
        language=request.language,
        prompt=PromptAgent(
            prompt=request.system_prompt,
            llm=request.llm, 
            temperature=0.5,
            tools=[
                {
                    "type": "system",
                    "description": "Converts text to speech using ElevenLabs API",
                    "name": "elevenlabs_tts",
                }
            ]
        ),
        first_message=request.first_message
### Additional context

  File "/Users/salmanaziz/Documents/twilio-bidirectional-rnd/venv/lib/python3.12/site-packages/pydantic/main.py", line 214, in __init__
    validated_self = self.__pydantic_validator__.validate_python(data, self_instance=self)
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
pydantic_core._pydantic_core.ValidationError: 3 validation errors for PromptAgent
tools.0.PromptAgentToolsItem_Webhook.type
  Input should be 'webhook' [type=literal_error, input_value='system', input_type=str]
    For further information visit https://errors.pydantic.dev/2.10/v/literal_error
tools.0.PromptAgentToolsItem_Webhook.api_schema.url
  Field required [type=missing, input_value={'type': 'object', 'prope...}, 'required': ['text']}, input_type=dict]
    For further information visit https://errors.pydantic.dev/2.10/v/missing
tools.0.PromptAgentToolsItem_Client.type
  Input should be 'client' [type=literal_error, input_value='system', input_type=str]
    For further information visit https://errors.pydantic.dev/2.10/v/literal_error

salman-aziz-4425:
Solution add this



<img width="1106" alt="Image" src="https://github.com/user-attachments/assets/81e7c21c-456e-44d8-b054-4f11f9218a01" />

PR #466, opened 2025-01-30
fix: system type validation in PromptAgent tool section

salman-aziz-4425:
[#464](https://github.com/elevenlabs/elevenlabs-python/issues/464)


salman-aziz-4425:
@louisjoecodes

PR #471, opened 2025-02-04
fix: make sure timeout is never None

### LINKED ISSUE #373: Timeout bug
Here's the type signature of the `httpx:AsyncClient.request` method:
```
    async def request(
        self,
        method: str,
        url: URL | str,
        *,
        content: RequestContent | None = None,
        data: RequestData | None = None,
        files: RequestFiles | None = None,
        json: typing.Any | None = None,
        params: QueryParamTypes | None = None,
        headers: HeaderTypes | None = None,
        cookies: CookieTypes | None = None,
        auth: AuthTypes | UseClientDefault | None = USE_CLIENT_DEFAULT,
        follow_redirects: bool | UseClientDefault = USE_CLIENT_DEFAULT,
        timeout: TimeoutTypes | UseClientDefault = USE_CLIENT_DEFAULT,
        extensions: RequestExtensions | None = None,
    ) -> Response:
```
It's important that they use a custom type to indicate that no timeout is set. When you wrap their client here:

https://github.com/elevenlabs/elevenlabs-python/blob/317b471eebfea68c70c35839a1d5c6264018446f/src/elevenlabs/core/client_wrapper.py#L63

You use `None` as the default timeout, and pass the default timeout value to get here:
https://github.com/elevenlabs/elevenlabs-python/blob/317b471eebfea68c70c35839a1d5c6264018446f/src/elevenlabs/core/http_client.py#L352

httpx treats `timeout=None` as no timeout at all, and this means that we're seeing some stuck connections in our code.

AngeloGiacco:
fixes #373

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ dist/
 __pycache__/
 poetry.toml
 .ruff_cache/
+env/
--- a/src/elevenlabs/base_client.py
+++ b/src/elevenlabs/base_client.py
@@ -93,7 +93,7 @@ class BaseElevenLabs:
         follow_redirects: typing.Optional[bool] = True,
         httpx_client: typing.Optional[httpx.Client] = None,
     ):
-        _defaulted_timeout = timeout if timeout is not None else 60 if httpx_client is None else float(httpx_client.timeout)
+        _defaulted_timeout = timeout if timeout is not None else 60 if httpx_client is None else httpx_client.timeout.read()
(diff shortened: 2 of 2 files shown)


louisjoecodes:
Hey @dsinghvi does it make sense to add this timeout flexibility upstream in Fern? 👀

AngeloGiacco:
is now resolved

PR #472, opened 2025-02-04
fix: fetch generated voice id and add to response

### LINKED ISSUE #345: ElevenLabs Python Client's voice_generation.generate does not return a generated_voice_id
Hi team,

elevenlabs==1.7.0

This is w.r.t. the python client's voice_generation.generate method which, at this point in time, returns a generator representing a stream of chunks that, when concatenated, result in a 'preview' mp3 of the generated voice.

The problem is that the return does not include the generated_voice_id which is necessary for re-using the voice.  Please review the APIs return type to include the necessary parameters.  According to the documentation, the generated_voice_id is returned as a response header which the client abstracts out.  

```python
xi = ElevenLabs(api_key='...')
xi_generated_voice = xi.voice_generation.generate(
    gender = 'female',
    age = 'young',
    accent = 'american',
    accent_strength=1.0,
    text = 'The quick brown fox jumps over the lazy dog.  Fifty-six crazy penguins jumped quickly over the awesome iceberg'
)
type(xi_generated_voice) # generator - where is the generated_voice_id?
```

Sincerest Regards,

AngeloGiacco:
currently we only return a bytes iterator with no reference to the voice id generated.
now return a tuple of the id string and then a bytes iterator, so sdk users can unpack and use as needed

fixes #345

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ dist/
 __pycache__/
 poetry.toml
 .ruff_cache/
+env/
--- a/src/elevenlabs/voice_generation/client.py
+++ b/src/elevenlabs/voice_generation/client.py
@@ -75,9 +75,11 @@ class VoiceGenerationClient:
         accent_strength: float,
         text: str,
         request_options: typing.Optional[RequestOptions] = None,
-    ) -> typing.Iterator[bytes]:
+    ) -> typing.Tuple[str, typing.Iterator[bytes]]:
         """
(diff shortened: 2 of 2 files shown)


louisjoecodes:
Will need to fix this one upstream as it will be overwritten
- is there anything you can tweak in openapi-overrides to replicate this diff?

louisjoecodes (changes requested):
Will need to fix this one upstream as it will be overwritten

is there anything you can tweak in openapi-overrides to replicate this diff?

Issue #368, opened 2024-09-30
BUG - Generate function streams output


lordsoffallen:
According the code snippet here: https://github.com/elevenlabs/elevenlabs-python/blob/main/src/elevenlabs/client.py#L225, that should run only when stream=False but following that line leads to here: https://github.com/elevenlabs/elevenlabs-python/blob/main/src/elevenlabs/text_to_speech/client.py#L122 which is where we client is being called via streaming endpoint and it actually yields, doesn't return. 

Must be bug in the code, as I expect that function to not stream but return the full output. Am I missing something here?

AngeloGiacco:
hi! thanks for the issue! our SDK has changed quite significantly since you raised this issue. can you confirm whether you are still having the issue, as the referenced code is out of date

AngeloGiacco:
confirmed this is an issue and have added a pr of a fix

PR #474, opened 2025-02-04
tts client when stream=false shouldnt stream

AngeloGiacco:
when stream is false, we call convert() as opposed to convert_as_stream(). convert before this PR also just streamed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ dist/
 __pycache__/
 poetry.toml
 .ruff_cache/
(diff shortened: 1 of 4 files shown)


AngeloGiacco:
will fix #368

louisjoecodes:
Amazing! This will for sure be overwritten so will keep this open so we can find a way to replicate this change

AngeloGiacco:
is now resolved

PR #358, opened 2024-09-11
(feat): Support for text iterators in `AsyncElevenLabs`

BackSlasher:
Redo of #346

Why:
Allowing the async client to utilize incoming text streams when generating voice. Very useful when feeding the realtime output of an LLM into the TTS.

Closes #344

What:
1. Copied `RealtimeTextToSpeechClient` and `text_chunker` into `AsyncRealtimeTextToSpeechClient` and `async_text_chunker` Most of the logic is intact, aside from async stuff
2. Added `AsyncRealtimeTextToSpeechClient` into `AsyncElevenLabs` just like `RealtimeTextToSpeechClient` is in `ElevenLabs`
3. Added rudimentary testing

The code is basically a copy-paste of what I found in the repo. We can rewrite it to be more elegant, but I figured parity with the sync code is more important.

PR #477, opened 2025-02-05
feat: asynciterable support

AngeloGiacco:
based off #358


AngeloGiacco:
@PaulAsjes this wouldnt be overwritten right?

PaulAsjes (changes requested):
This won't be overwritten by autogen, which is why we need way more tests here imo.

PR #479, opened 2025-02-10
fix: extract request-id from response headers

### LINKED ISSUE #478: Add response attribute to TextToSpeechClient class
### Feature Description

Ability to query the response headers, specifically to get the request-id.

The client.TextToSpeechClient class doesn't expose the HTTPS response object, nor the request-id. That makes it difficult to populate the `previous_request_ids` argument to methods such as `convert`.

Including the response as an attribute would allow generational conditioning from the Python API, as in the code below:

```
previous_request_ids = deque(maxlen=MAX_REQUEST_IDS)

# The following code is in a loop iterating over the lines of text.

try:
    previous_text = OMIT if is_first_paragraph or len(text) == 1 else ' '.join(text[:line_no])
    next_text = OMIT if is_last_paragraph or len(text) == 1 else ' '.join(text[line_no + 1:])

    previous_request_argument = previous_request_ids if ((len(previous_request_ids) > 0)
                                                         and condition_on_past_generations) else OMIT

    audio = client.text_to_speech.convert(
        voice_id=voice_id,
        optimize_streaming_latency=None,
        output_format=output_format,
        text=line,
        model_id=model,
        voice_settings=settings,
        previous_text=previous_text,
        next_text=next_text,
        previous_request_ids=previous_request_argument,
        pronunciation_dictionary_locators=pronunciation_dictionary_locators
    )
except Exception as e:
    log_exception(e)
    print(e)
    break
else:
    for chunk in audio:
        audio_segments.append(chunk)
    if condition_on_past_generations:
        request_id = client.text_to_speech.response.headers['request-id']    # <------------------
        if request_id:
            previous_request_ids.append(request_id)
```

### Use Case

Passing previous request IDs to the various methods of the TextToSpeechClient object.

The methods accept a `previous_request_ids` argument, but the class doesn't provide an obvious way to obtain the request IDs.

### Alternatives Considered

At the moment I edit client.py to add the attribute:

```
class TextToSpeechClient:
    def __init__(self, *, client_wrapper: SyncClientWrapper):
        self._client_wrapper = client_wrapper
        self.response = None
```

then update self.response from _response in each of the methods. I have to make that change again each time the elevenlabs package is updated.

### Additional Context

_No response_

AngeloGiacco:
fixes #478

### DIFF
--- a/src/elevenlabs/text_to_speech/client.py
+++ b/src/elevenlabs/text_to_speech/client.py
@@ -61,9 +61,9 @@ class TextToSpeechClient:
             BodyTextToSpeechV1TextToSpeechVoiceIdPostApplyTextNormalization
         ] = OMIT,
         request_options: typing.Optional[RequestOptions] = None,
-    ) -> typing.Iterator[bytes]:
+    ) -> tuple[str, typing.Iterator[bytes]]:
         """
-        Converts text into speech using a voice of your choice and returns audio.
+        Converts text into speech using a voice of your choice and returns the request ID and audio stream.
 
         Parameters
         ----------
@@ -126,9 +126,11 @@ class TextToSpeechClient:
             Request-specific configuration. You can pass in configuration such as `chunk_size`, and more to customize the request and response.
 
         Yields
-        ------
-        typing.Iterator[bytes]
-            Successful Response
+        -------
+        tuple[str, typing.Iterator[bytes]]
+            A tuple containing:
+            - request_id: The ID of the request
+            - audio_stream: Iterator of audio bytes chunks
 
         Examples
         --------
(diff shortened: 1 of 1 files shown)


src/elevenlabs/text_to_speech/client.py
```diff
@@ -180,10 +182,20 @@ def convert(
         ) as _response:
             try:
                 if 200 <= _response.status_code < 300:
-                    _chunk_size = request_options.get("chunk_size", 1024) if request_options is not None else 1024
-                    for _chunk in _response.iter_bytes(chunk_size=_chunk_size):
-                        yield _chunk
-                    return
+                    request_id = _response.headers.get('request-id')
+                    if not request_id:
+                        raise ApiError(
```
PaulAsjes, line 187: How likely is this to happen?

PaulAsjes (approved):
Looks good, just one little nit.

src/elevenlabs/text_to_speech/client.py
AngeloGiacco: impossible because we set it if not provided by the user

AngeloGiacco:
please don't merge, want to leave for fern folks to see

PR #482, opened 2025-02-12
Fix: Error receiving message & Error sending user audio chunk sent 1009 (message too big)

nextgLoki:
Sometimes using the ConversationalAI [demo](https://github.com/elevenlabs/elevenlabs-examples/blob/main/examples/conversational-ai/python/convai/demo.py) throwing exceptions like:
- `Error receiving message: sent 1009 (message too big) frame with 1254313 bytes exceeds limit of 1048576 bytes; no close frame received`
- `Error sending user audio chunk: sent 1009 (message too big) frame with 1254313 bytes exceeds limit of 1048576 bytes; no close frame received`

Seems current ws message limit is too low for some provided sound answers. This fix should fits long response requirements.

### DIFF
--- a/src/elevenlabs/conversational_ai/conversation.py
+++ b/src/elevenlabs/conversational_ai/conversation.py
@@ -272,7 +272,7 @@ class Conversation:
         return self._conversation_id
 
     def _run(self, ws_url: str):
-        with connect(ws_url) as ws:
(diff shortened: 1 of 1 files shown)


AngeloGiacco:
haha nice catch, i found this same bug in an external user of elevenlabs ws two weeks ago, and now its in our own sdk. annoying. thank you!

AngeloGiacco (approved):

Issue #485, opened 2025-02-13
align Python SDK to API for ConversationConfig


futilm:
### Description

Hello,

I noticed that the Conversation module in this Python SDK may not be aligned with the API.

This occurs for `ConversationConfig` 

https://github.com/elevenlabs/elevenlabs-python/blob/447362d4e381310a8119a88c7c3ca98f914d3393/src/elevenlabs/conversational_ai/conversation.py#L166-L174

used for example in the instanciation of `ConversationalConfig` 

https://github.com/elevenlabs/elevenlabs-python/blob/447362d4e381310a8119a88c7c3ca98f914d3393/src/elevenlabs/types/conversational_config.py#L19-L24

which is not the same as the API that accepts attributes `max_duration_seconds` and `client_events` (https://elevenlabs.io/docs/conversational-ai/api-reference/agents/create-agent#request.body.conversation_config.conversation).

Moreover, the properties currently used in `ConversationConfig` appears to be the ones used in the `platform_settings` object (`conversation_config_override` and `custom_llm_extra_body`): https://elevenlabs.io/docs/conversational-ai/api-reference/agents/create-agent#request.body.platform_settings.overrides

Probably the correct class to be used is

https://github.com/elevenlabs/elevenlabs-python/blob/447362d4e381310a8119a88c7c3ca98f914d3393/src/elevenlabs/types/conversation_config.py#L10-L12

which is the one aligned with the API that in fact has properties `max_duration_seconds` and `client_events`.

Is that a name clash or am I missing something?

Thank you

AngeloGiacco:
hi thanks for the issue! i agree that the naming is a little confusing. 

i don't get an issue with the code though, because

https://github.com/elevenlabs/elevenlabs-python/blob/447362d4e381310a8119a88c7c3ca98f914d3393/src/elevenlabs/conversational_ai/conversation.py#L166-L174

 is correctly referring to

https://github.com/elevenlabs/elevenlabs-python/blob/447362d4e381310a8119a88c7c3ca98f914d3393/src/elevenlabs/types/conversation_config.py#L10-L12 . 

It is not importing https://github.com/elevenlabs/elevenlabs-python/blob/447362d4e381310a8119a88c7c3ca98f914d3393/src/elevenlabs/conversational_ai/conversation.py#L166-L174 like you mentioned. 

could you double check?  

[...]

PR #486, opened 2025-02-13
refactor: renaming to remove clash

AngeloGiacco:
based on #485


PaulAsjes:
Won't this be rewritten by fern?

PaulAsjes (approved):
Double checked this this won't be rewritten 👍

PR #514, opened 2025-03-30
added next and previous text as arguments for the generate functions,…

aalghooneh:
… so one can enforce emotions

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,6 @@
+{
+    "workbench.colorCustomizations": {
+        "minimap.background": "#00000000",
+        "scrollbar.shadow": "#00000000"
+    }
+}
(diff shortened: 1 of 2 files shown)


aalghooneh:
@flavioschneider I cam across this use case in my application, and it seems this part is missing in the python version, this was already in the JS version of this library (Check this [video](https://www.youtube.com/watch?v=ZFdW--YfjjM) to see its impact!) give it a look

aalghooneh:
@PaulAsjes

PaulAsjes:
Hi @aalghooneh, thanks for the contribution, but we're actually going to be deprecating and removing the generate function soon. Instead you should use the underlying `client.text_to_speech.convert_as_stream` and `client.text_to_speech.convert` methods.

aalghooneh:
thanks Paul, good to hear that

Issue #519, opened 2025-04-05
Documentation example on adding client tool returns exception


BonifacioCalindoro:
### Description

with latest pip release of elevenlabs python sdk (1.56.0)

**Steps to reproduce**

1. Create a client-side tool from the elevenlabs dashboard as mentioned in the [client-tools documentation](https://elevenlabs.io/docs/conversational-ai/customization/tools/client-tools)
2. Run the example code for registering client-tools with the appropiate api_key and agent_id that is mentioned in the documentation

**Expected behavior**
- Basically that it works as intended with the instructions provided by the documentation.

**Actual behavior**
- Returns an exception:
```python

Traceback (most recent call last):
  File "[redacted]/elevenlabs_client.py", line 15, in <module>
    conversation = Conversation(
                   ^^^^^^^^^^^^^
TypeError: Conversation.__init__() missing 2 required keyword-only arguments: 'requires_auth' and 'audio_interface'`

### Code example

```python

from elevenlabs import ElevenLabs
from elevenlabs.conversational_ai.conversation import Conversation, ClientTools
import os
from dotenv import load_dotenv

load_dotenv()

def log_message(parameters):
    message = parameters.get("message")
    print(message)

client_tools = ClientTools()
client_tools.register("logMessage", log_message)

conversation = Conversation(
    client=ElevenLabs(api_key=os.getenv('ELEVENLABS_API_KEY')),
    agent_id=os.getenv('ELEVENLABS_AGENT_ID'),
    client_tools=client_tools,
    # ...
)

conversation.start_session()
```

### Additional context

When setting requires_auth=False and audio_interfacte=None, i get this other exception:

```python

Traceback (most recent call last):
  File "[redacted]/elevenlabs_client.py", line 24, in <module>
    conversation.start_session()
  File "[redacted]/.venv/lib/python3.12/site-packages/elevenlabs/conversational_ai/conversation.py", line 252, in start_session
    ws_url = self._get_signed_url() if self.requires_auth else self._get_wss_url()
                                                               ^^^^^^^^^^^^^^^^^^^
  File "[redacted]/.venv/lib/python3.12/site-packages/elevenlabs/conversational_ai/conversation.py", line 373, in _get_wss_url
    base_url = self.client._client_wrapper._base_url
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'SyncClientWrapper' object has no attribute '_base_url'`

[...]

PR #526, opened 2025-04-14
Fix: Update Conversation WSS URL Construction to Use Latest SyncClientWrapper Implementation

JuanEgas:
Description:
This PR updates the _get_wss_url method in the Conversation object to correctly construct the WebSocket URL when require_auth is set to False. Previously, the URL construction was based on an older version of SyncClientWrapper, which resulted in AttributeError: 'SyncClientWrapper' object has no attribute '_base_url'`

Current version (Old implementation): 
Screenshot 2025-04-14 at 12 01 41 PM

Updated Version: 
Screenshot 2025-04-14 at 12 00 48 PM

> Refactored _get_wss_url method:
Now retrieves the base URL using the latest version of SyncClientWrapper. 
> Introduced New Test Case:
Added tests to validate the correct construction of WSS URLs under different base URL scenarios (both HTTPS and HTTP). 

Related issues: 
https://github.com/elevenlabs/elevenlabs-python/issues/519

### DIFF
--- a/src/elevenlabs/conversational_ai/conversation.py
+++ b/src/elevenlabs/conversational_ai/conversation.py
@@ -370,7 +370,7 @@ class Conversation:
             pass  # Ignore all other message types.
 
     def _get_wss_url(self):
-        base_url = self.client._client_wrapper._base_url
+        base_url = self.client._client_wrapper.get_environment().base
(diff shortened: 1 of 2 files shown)


AngeloGiacco:
closing this in favour of #542

PR #578, opened 2025-06-27
feat: add callback for session end in Conversation class

8thgencore:
* Added `callback_end_session` parameter to the Conversation class to allow users to define a callback function that will be executed when the session ends.
* Updated the `__init__` method and `end_session` method to handle the new callback.

This enhancement provides more flexibility for managing session termination events.


AngeloGiacco (approved):
lgtm

PR #587, opened 2025-07-08
feat: add user_id to convai

jameszhou02:


AngeloGiacco (approved):
lgtm, but can we keep diff smaller in future, hard to read

PaulAsjes (approved):

Issue #708, opened 2025-12-26
`text_to_speech.convert` of `AsyncElevenLabs` returns unexpected type which cannot be awaited


legit-programmer:
**To reproduce:**
```
client = AsyncElevenLabs(
  api_key=os.getenv("ELEVENLABS_API_KEY"),
)


async def main() -> None:
    await client.text_to_speech.convert(
        voice_id="rAsfH6d68tmh0XRGXp4D",
        output_format="mp3_44100_128",
        text="The first move is what sets everything in motion.",
        model_id="eleven_multilingual_v2",
    )


asyncio.run(main())
```
**Error on console:**
`TypeError: object async_generator can't be used in 'await' expression`

kraenhansen:
Thanks for reporting this, @legit-programmer. The async `convert()` method returns an `AsyncIterator[bytes]` (an async generator), so it should be iterated with `async for` rather than `await`ed. Here's the corrected approach:

```python
async def main() -> None:
    audio_chunks = []
    async for chunk in client.text_to_speech.convert(
        voice_id="rAsfH6d68tmh0XRGXp4D",
        output_format="mp3_44100_128",
        text="The first move is what sets everything in motion.",
        model_id="eleven_multilingual_v2",
    ):
        audio_chunks.append(chunk)
    audio = b"".join(audio_chunks)
```

This was also covered in #618. I'm closing this as a duplicate.

PR #709, opened 2025-12-29
Fix Async TTS convert() documentation and add awaitable convert_to_bytes helper #708

aijadugar:
Problem:
AsyncTextToSpeechClient.convert() is implemented as an async generator but is documented and demonstrated as an awaitable coroutine, causing a runtime

TypeError for users. #708 

Solution:
- Corrected examples to reflect streaming semantics (async for)
- Added convert_to_bytes() as a convenience helper for users who want
  an awaitable API returning full audio bytes
- Preserved backward compatibility
- Added tests to lock expected behavior

tests/async_eleven.py =>

image

### 🔊 Audio Sample
🎧 **Listen:** [test.mp3](https://github.com/user-attachments/files/24372359/test.mp3)

### DIFF
--- a/src/elevenlabs/text_to_speech/client.py
+++ b/src/elevenlabs/text_to_speech/client.py
@@ -588,6 +588,26 @@ class AsyncTextToSpeechClient:
         """
         return self._raw_client
 
+    async def convert_to_bytes(
+        self,
+        voice_id: str,
+        *,
+        text: str,
+        **kwargs,
+    ) -> bytes:
+        """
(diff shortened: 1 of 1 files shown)


PaulAsjes:
Thanks, but we won't be able to merge this as it makes changes to autogenerated files. Your changes would be overwritten the next time we generate the SDKs. Could instead open up an issue detailing exactly what the problem is? We can then forward that to our SDK provider for a fix.

PR #713, opened 2026-01-12
[Agents] Add audio alignment callback

alexgoodison:
This PR introduces the ability to handle audio alignment data sent as part of an agents conversation by adding a callback to the Conversation class.

### DIFF
--- a/src/elevenlabs/conversational_ai/conversation.py
+++ b/src/elevenlabs/conversational_ai/conversation.py
@@ -2,6 +2,7 @@ from abc import ABC, abstractmethod
 import asyncio
 import base64
 from concurrent.futures import ThreadPoolExecutor
+from dataclasses import dataclass
 from enum import Enum
 import json
 import logging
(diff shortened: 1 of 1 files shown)


src/elevenlabs/conversational_ai/conversation.py
```diff
@@ -364,12 +375,16 @@ def __init__(
     def _get_wss_url(self):
         if self.on_prem_config:
             return self.on_prem_config.on_prem_conversation_url
-            
+
         base_http_url = self.client._client_wrapper.get_base_url()
-        base_ws_url = urllib.parse.urlparse(base_http_url)._replace(scheme="wss" if base_http_url.startswith("https") else "ws").geturl()
```
AngeloGiacco: lets avoid formatting changes, makes PR harder to parse

alexgoodison: yeah fair, will undo

Issue #732, opened 2026-02-19
Music client is overwritten by music_custom.py excluding some arguments


groadabike:
### Description

Hi, 
When using `client.music.composed_detailed` method and trying to set the parameter `with_timestamps` to True, you get a TypeError message
`TypeError: AsyncMusicClient.compose_detailed() got an unexpected keyword argument 'with_timestamps'`

However, the argument should be valid, as it is documented in https://github.com/elevenlabs/elevenlabs-python/blob/0b87e7712b1e84957aae4e45f89e9c29d45eb4ba/src/elevenlabs/music/client.py#L414

The problem appears to be that the default client is overwritten by https://github.com/elevenlabs/elevenlabs-python/blob/0b87e7712b1e84957aae4e45f89e9c29d45eb4ba/src/elevenlabs/music_custom.py at runtime, which doesn't have that parameter.

### Code example

```python
client = ElevenLabs(api_key=api_key)

audio_generator = client.music.compose_detailed(
    prompt='prompt',
    output_format='pcm_44100',
    music_length_ms=30000,
    with_timestamps=True
)
```

PR #733, opened 2026-02-19
[Music] Add missing music parameters

PaulAsjes:
Fixes https://github.com/elevenlabs/elevenlabs-python/issues/732

TODO: We need a way of automatically adding these.


PaulAsjes:
Great idea on the `**kwargs`!

PR #771, opened 2026-04-14
Speech Engine SDK

PaulAsjes:
> [!NOTE]
> **Medium Risk**
> Adds a new async WebSocket server/session layer plus JWT verification for inbound Speech Engine connections and bumps the `websockets` dependency, so runtime behavior and compatibility may change even though changes are largely additive.
> 
> **Overview**
> Introduces a new `speech_engine` module enabling server-side voice agents: a `SpeechEngineServer` (standalone WebSocket server) and `SpeechEngineSession` (event-emitter style session that streams `agent_response` messages, supports interruption/cancellation, and auto-parses common LLM streaming formats).
> 
> Extends the generated Speech Engine clients with `SpeechEngineResource` wrappers returned from `create`/`get`/`update`, adds `speech_engine` accessors to `ElevenLabs`/`AsyncElevenLabs`, and documents usage in `README.md`.
> 
> Adds HS256 JWT request verification for inbound connections (used by `SpeechEngineServer` and exposed via `SpeechEngineResource.verify_request`), comprehensive new tests for auth/session/server behavior, updates `.fernignore` to protect the new module, and bumps `websockets` to `>=13.0`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit dc02e44abf13ed30e1712b12881a0d6d1eabc787. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/elevenlabs/speech_engine/resource.py
```diff
@@ -0,0 +1,180 @@
...
+        )
+    """
+
+    def __init__(
+        self,
+        engine_id: str,
+        client_options: typing.Any = None,
+    ) -> None:
+        self.engine_id = engine_id
+        self._options = client_options
```
kraenhansen: **Fragile API key access pattern**

This accesses the client wrapper's private attribute directly:

```python
def _get_api_key(self) -> typing.Optional[str]:
    if self._options is not None and hasattr(self._options, "_api_key"):
        return self._options._api_key
    return None
```

This is fragile — if the internal client wrapper ever renames `_api_key`, this will silently return `None` and all request verification will fail (returning `False` without any error).

For comparison, the JS SDK uses the proper supplier pattern: `await core.Supplier.get(this._options.apiKey)`.

Consider accessing the API key through a more stable interface, or at minimum raising an error if the expected attribute isn't found rather than silently returning `None`.

src/elevenlabs/speech_engine/server.py
```diff
@@ -0,0 +1,121 @@
...
+
+        self._stop_event = asyncio.Event()
+
+        async def _handler(websocket: typing.Any, *_args: typing.Any) -> None:
+            if self._path is not None and websocket.request.path != self._path:
+                await websocket.close(4000, "not found")
+                return
+
+            header_value = websocket.request.headers.get(
+                "x-elevenlabs-speech-engine-authorization"
```
kraenhansen, line 93: **Security: Auth check happens after WebSocket handshake completes**

Currently, unauthenticated requests complete the full WebSocket handshake before being rejected via `websocket.close()`. In contrast, the [JS SDK equivalent](https://github.com/elevenlabs/elevenlabs-js/pull/365) rejects at the HTTP level before upgrading — unauthenticated clients never get a WebSocket connection.

The `websockets` library (v11+) supports a `process_request` callback on `serve()` that runs **before** the handshake. This would let you reject at the HTTP level, matching the JS behavior:

```python
import http

def _check_request(connection, request):
    if self._path is not None and request.path != self._path:
        return connection.respond(http.HTTPStatus.NOT_FOUND, "not found\n")

    header_value = request.headers.get(
        "x-elevenlabs-speech-engine-authorization"
    )
    if not header_value:
        self._log(
            "rejected connection — missing "
            "X-Elevenlabs-Speech-Engine-Authorization header"
        )
        return connection.respond(
            http.HTTPStatus.UNAUTHORIZED,
            "missing authorization header\n",
        )
    try:
        verify_speech_engine_jwt(header_value, api_key)
    except ValueError as e:
        self._log("rejected connection — %s", e)
        return connection.respond(
            http.HTTPStatus.UNAUTHORIZED,
            "authorization failed\n",
        )
    return None  # accept — proceed with WebSocket handshake

async def _handler(websocket):
    self._log("verified connection, accepting WebSocket")
    session = self.handle_connection(websocket)
    await session.run()

self._server = await websockets.serve(
    _handler, "", self._port,
    process_request=_check_request,
)
```

This way unauthenticated requests get a plain HTTP 401 without ever upgrading to WebSocket — less resource consumption, no info leakage via close frames.

src/elevenlabs/client.py
```diff
@@ -64,6 +64,24 @@ def __init__(
...
+    def __init__(self, client_wrapper: typing.Any) -> None:
+        self._client_wrapper = client_wrapper
+
+    async def get(self, engine_id: str) -> "SpeechEngineResource":
+        from .speech_engine.resource import SpeechEngineResource  # noqa: E402
+
+        return SpeechEngineResource(
+            engine_id=engine_id,
+            client_options=self._client_wrapper,
+        )
```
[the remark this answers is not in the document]
PaulAsjes: Fixed — renamed the parameter from `client_options` to `client_wrapper` in both `SpeechEngineResource.__init__()` and all call sites (`client.py`, tests).

src/elevenlabs/speech_engine/server.py
```diff
@@ -0,0 +1,75 @@
...
+        self,
+        *,
+        port: int = 3001,
+        debug: bool = False,
+        **handlers: typing.Any,
+    ) -> None:
+        self._port = port
+        self._debug = debug
+        self._handlers = handlers
+        self._stop_event = asyncio.Event()
```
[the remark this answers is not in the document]
PaulAsjes: Fixed — `self._stop_event` is now initialized to `None` in `__init__` and only created as `asyncio.Event()` inside `serve()` (which runs within an event loop).

src/elevenlabs/speech_engine/session.py
```diff
@@ -0,0 +1,479 @@
...
+        self._ws = ws
+        self._conversation_id = None  # type: typing.Optional[str]
+        self._current_task = None  # type: typing.Optional[asyncio.Task]  # type: ignore[type-arg]
+        self._current_event_id = None  # type: typing.Optional[int]
+        self._closed = False
+        self._event_handlers = {}  # type: typing.Dict[str, typing.List[Callback]]
+        self._once_handlers = {}  # type: typing.Dict[str, typing.List[Callback]]
+
+        if debug:
+            logger.setLevel(logging.DEBUG)
```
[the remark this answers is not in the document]
PaulAsjes: Fixed — replaced `logger.setLevel(logging.DEBUG)` with a per-instance `_make_log(debug)` function that uses `print()` when debug is on and a no-op lambda when off. No global logger mutation.

src/elevenlabs/speech_engine/session.py
```diff
@@ -0,0 +1,479 @@
...
+            all_handlers = handlers + once_handlers
+
+            if all_handlers:
+                self._current_task = asyncio.create_task(
+                    self._run_transcript_handlers(all_handlers, transcript)
+                )
+                # Yield so the handler task can start before the next
+                # message is read.  This mirrors the JS behaviour where
+                # emitter.emit() invokes listeners synchronously.
+                await asyncio.sleep(0)
```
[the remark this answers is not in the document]
PaulAsjes: By design — `once` handlers are consumed (popped) when the event fires, regardless of whether the handler completes or is cancelled. This matches the semantics of EventEmitter `once()` in Node.js: "fire at most one time" means it is removed on the first fire, not on completion. Re-registering a `once` handler after cancellation is straightforward if needed.

src/elevenlabs/speech_engine/session.py
```diff
@@ -0,0 +1,479 @@
...
+            logger.debug(
+                "received transcript, event_id=%s, messages=%d",
+                self._current_event_id,
+                len(transcript_data),
+            )
+
+            transcript = [
+                ConversationMessage(role=m["role"], content=m["content"])
+                for m in transcript_data
+            ]
```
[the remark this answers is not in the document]
PaulAsjes: Fixed — the transcript parsing is now wrapped in `try/except (KeyError, TypeError)` which emits an error event instead of crashing the session.

src/elevenlabs/speech_engine/server.py
```diff
@@ -0,0 +1,129 @@
...
+                await websocket.close(
+                    4001, "missing authorization header"
+                )
+                return
+
+            try:
+                verify_speech_engine_jwt(header_value, api_key)
+            except ValueError as e:
+                logger.debug("rejected connection — %s", e)
+                await websocket.close(4001, str(e))
```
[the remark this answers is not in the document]
PaulAsjes: Fixed — the close message sent to clients is now a generic `"authorization failed"` string instead of the raw `ValueError` message, so no API key details are leaked.

src/elevenlabs/speech_engine/session.py
```diff
@@ -0,0 +1,506 @@
...
+    ) -> None:
+        try:
+            for handler in handlers:
+                result = handler(transcript)
+                if asyncio.iscoroutine(result):
+                    await result
+        except asyncio.CancelledError:
+            raise
+        except Exception as e:
+            await self._emit("error", e)
```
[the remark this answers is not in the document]
PaulAsjes: By design — the current behavior (single try/except around the handler loop) is intentional. When a transcript arrives, the first handler that fails should stop processing for that transcript since subsequent handlers would be operating on potentially stale context. Isolating handler errors would mask failures and make debugging harder. Errors are caught and logged at the task level, so the session continues processing future transcripts.

src/elevenlabs/speech_engine/session.py
```diff
@@ -0,0 +1,513 @@
...
+            if (
+                incoming_event_id == self._current_event_id
+                and self._current_task is not None
+                and not self._current_task.done()
+            ):
+                self._log(
+                    "skipping duplicate transcript, event_id=%s",
+                    incoming_event_id,
+                )
+                return
```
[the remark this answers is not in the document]
PaulAsjes: Fixed — the deduplication guard now checks `incoming_event_id is not None` before comparing, so `None` event IDs are never treated as duplicates.

src/elevenlabs/speech_engine/resource.py
PaulAsjes: Fixed — `_get_api_key()` now uses the public `get_headers()` method on the client wrapper instead of accessing private attributes directly:

```python
def _get_api_key(self) -> typing.Optional[str]:
    if self._client_wrapper is None:
        return None
    headers = self._client_wrapper.get_headers()
    return headers.get("xi-api-key")
```

src/elevenlabs/speech_engine/server.py
PaulAsjes: Acknowledged — this is a known limitation of the `websockets` library, which does not support rejecting connections before the handshake completes (there is no HTTP-level hook). The auth check happens immediately after the handshake, so unauthenticated connections are closed within milliseconds. For the standalone `serve()` server this is acceptable. For production deployments where pre-handshake rejection matters, we recommend using `create_session()` with a framework like FastAPI/Starlette where you can inspect headers before accepting the WebSocket upgrade.

src/elevenlabs/speech_engine/session.py
```diff
@@ -0,0 +1,514 @@
...
+            raise RuntimeError("Cannot send response: session is closed")
+
+        if self._current_event_id is None:
+            logger.warning(
+                "sendResponse() called outside of an on_transcript handler. "
+                "Responses can only be sent in reply to a user transcript. "
+                "To have the agent speak first, set a first message in your "
+                "Speech Engine conversation config on the client."
+            )
+            return
```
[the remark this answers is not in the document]
PaulAsjes: Good catch — fixed. The guard now uses a dedicated `_in_transcript_handler` flag instead of checking `_current_event_id is None`, so transcripts without an `event_id` work correctly.

src/elevenlabs/speech_engine/session.py
```diff
@@ -0,0 +1,518 @@
...
+                    incoming_event_id,
+                )
+
+            self._current_event_id = incoming_event_id
+            transcript_data = msg.get("user_transcript", [])
+            self._log(
+                "received transcript, event_id=%s, messages=%d",
+                self._current_event_id,
+                len(transcript_data),
+            )
```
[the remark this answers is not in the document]
PaulAsjes: Good catch — fixed. `msg.get("user_transcript")` returns `None` when the key exists with a null value, and `len(None)` crashes before the try/except. Fixed by using `or []` to coerce null to an empty list, and moved the log call after the validation block.

src/elevenlabs/speech_engine/server.py
kraenhansen: And you didn't want to adopt the approach of passing `process_request`? Would it be more complicated for no apparent benefit? It seems valuable to me to close the connection as soon as possible, if the token is invalid 🤔

src/elevenlabs/speech_engine/server.py
```diff
@@ -0,0 +1,121 @@
...
+        self._stop_event = asyncio.Event()
+
+        async def _handler(websocket: typing.Any, *_args: typing.Any) -> None:
+            if self._path is not None and websocket.request.path != self._path:
+                await websocket.close(4000, "not found")
+                return
+
+            header_value = websocket.request.headers.get(
+                "x-elevenlabs-speech-engine-authorization"
+            )
```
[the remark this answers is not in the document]
PaulAsjes: Valid — the server uses `websocket.request.path` and `websocket.request.headers` which are websockets 13.x API. Bumped the minimum from `>=11.0` to `>=13.0` in `pyproject.toml`. Since the speech engine feature is new, there's no backwards compatibility concern.

src/elevenlabs/speech_engine/session.py
```diff
@@ -0,0 +1,519 @@
...
+                # message is read.  This mirrors the JS behaviour where
+                # emitter.emit() invokes listeners synchronously.
+                await asyncio.sleep(0)
+
+        elif msg_type == "ping":
+            await self._send({"type": "pong"})
+
+        elif msg_type == "close":
+            await self._cancel_current_and_wait()
+            await self._emit("close")
```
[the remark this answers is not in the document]
PaulAsjes: Valid — the close handler now sets `self._closed = True` before cancelling tasks and emitting the event, so the `run()` loop exits and `is_open` returns `False`.

src/elevenlabs/speech_engine/resource.py
```diff
@@ -0,0 +1,181 @@
...
+
+def _base64url_decode(data: str) -> bytes:
+    padded = data.replace("-", "+").replace("_", "/")
+    remainder = len(padded) % 4
+    if remainder:
+        padded += "=" * (4 - remainder)
+    return base64.b64decode(padded)
+
+
+def verify_speech_engine_jwt(value: str, api_key: str) -> typing.Dict[str, typing.Any]:
```
kraenhansen, line 30: I'm wondering if you've considered using something less custom for this, like [PyJWT](https://pypi.org/project/PyJWT/)?

src/elevenlabs/speech_engine/session.py
```diff
@@ -0,0 +1,521 @@
...
+            return
+
+        if isinstance(response, str):
+            self._log(
+                'sending string response: "%s", event_id=%s',
+                response,
+                self._current_event_id,
+            )
+            await self._send_agent_response(response, False)
+            await self._send_agent_response("", True)
```
kraenhansen: Curious if / why this extra send is needed 🤔

src/elevenlabs/speech_engine/session.py
```diff
@@ -0,0 +1,521 @@
...
+            handlers.remove(handler)
+        except ValueError:
+            pass
+        return self
+
+    def once(self, event: str, handler: Callback) -> "SpeechEngineSession":
+        """Register *handler* for *event*, removed after the first call."""
+        self._once_handlers.setdefault(event, []).append(handler)
+        return self
+
```
kraenhansen, line 209: I'm slightly concerned / curious to learn if this smells a bit too much like a port of a JavaScript API? I don't have a strong signal here.

src/elevenlabs/speech_engine/session.py
```diff
@@ -0,0 +1,521 @@
...
+                result = handler(*args)
+                if asyncio.iscoroutine(result):
+                    await result
+            except asyncio.CancelledError:
+                raise
+            except Exception as e:
+                if event != "error":
+                    await self._emit("error", e)
+                else:
+                    logger.exception("unhandled error in error handler")
```
kraenhansen: Would it make sense to propagate the exception too?

```suggestion
                    logger.exception("unhandled error in error handler", e)
```

src/elevenlabs/speech_engine/session.py
```diff
@@ -0,0 +1,521 @@
...
+            raise RuntimeError("Cannot send response: session is closed")
+
+        if not self._in_transcript_handler:
+            logger.warning(
+                "send_response() called outside of an on_transcript handler. "
+                "Responses can only be sent in reply to a user transcript. "
+                "To have the agent speak first, set a first message in your "
+                "Speech Engine conversation config on the client."
+            )
+            return
```
kraenhansen, line 301: Feels like a bit of an arbitrary restriction, is this a server-side constraint or an API design choice?

src/elevenlabs/speech_engine/session.py
```diff
@@ -0,0 +1,521 @@
...
+                    self._current_event_id,
+                    incoming_event_id,
+                )
+
+            self._current_event_id = incoming_event_id
+            transcript_data = msg.get("user_transcript") or []
+
+            try:
+                transcript = [
+                    ConversationMessage(role=m["role"], content=m["content"])
```
kraenhansen, line 380: We might want to be more defensive here, to fail faster in case the data from the server is malformed.

src/elevenlabs/speech_engine/session.py
```diff
@@ -0,0 +1,521 @@
...
+            asyncio.ensure_future(self._ws.close())
+        except RuntimeError:
+            # No running event loop — best-effort.
+            pass
+
+    # ------------------------------------------------------------------
+    # Internals
+    # ------------------------------------------------------------------
+
+    async def _handle_message(self, msg: typing.Dict[str, typing.Any]) -> None:
```
kraenhansen, line 336: If this was TypeScript code I'd suggest not using `Any` anywhere to force "downstream" type checking before code start relying on some implicit type. I'd say that applies here as well.

```suggestion
    async def _handle_message(self, msg: object) -> None:
```

src/elevenlabs/speech_engine/session.py
```diff
@@ -0,0 +1,521 @@
...
+                    raw = await self._ws.recv()
+                except asyncio.CancelledError:
+                    raise
+                except Exception:
+                    self._log("WebSocket connection lost")
+                    break
+
+                try:
+                    if isinstance(raw, bytes):
+                        raw = raw.decode("utf-8")
```
kraenhansen, line 250: Could we assert here instead of covering two potential return values from `recv`? I _suspect_ we'd never expect a binary message from the server as is.

kraenhansen (approved):
I have a few more comments 🙈 Good to merge as-is 👍

src/elevenlabs/speech_engine/resource.py
PaulAsjes: This implementation is only ~40 lines, I think the tradeoff of not having another dependency is worth it.

src/elevenlabs/speech_engine/session.py
PaulAsjes: Server-side constraint. If you attempt to send a response out of turn it gets ignored by the API. Better to warn about that instead of silently fail.

src/elevenlabs/speech_engine/session.py
PaulAsjes: The websockets library's `recv()` return type is `str | bytes` — text frames return `str` and binary frames return bytes. In practice the Speech Engine API only sends text frames, but decoding bytes as UTF-8 is a one-line defensive fallback that keeps the session alive if a binary frame somehow arrives. An assert would crash the session instead. I'd rather keep the graceful handling.

src/elevenlabs/speech_engine/session.py
PaulAsjes: The message is always a parsed JSON object from `json.loads()`, and `_handle_message` calls dict methods like `msg.get("type")` and `msg.get("user_transcript")` throughout. Using `object` would require explicit type narrowing or casts at every access point. `Dict[str, Any]` is the accurate type here.

src/elevenlabs/speech_engine/session.py
PaulAsjes: The transcript parsing is wrapped in try/except (KeyError, TypeError) which emits an error event instead of crashing the session. The `transcript_data` assignment also uses `or []` to handle null values.

src/elevenlabs/speech_engine/session.py
PaulAsjes: The on/off/once event emitter pattern is intentional — it keeps the API consistent across both SDKs so developers switching between languages encounter the same concepts. The pattern is also well-established in Python via libraries like `pyee` and `asyncio`'s own event handling. The alternative (subclassing / abstract methods) would make the session harder to compose and wouldn't support multiple handlers per event

src/elevenlabs/speech_engine/session.py
PaulAsjes: Wire protocol requirement. The empty message with `is_final: True` signals to the Speech Engine API that the response is complete and TTS synthesis should finalize. Without it, the API would keep waiting for more content. This matches the streaming case where chunks are sent with `is_final: False` and the final empty message terminates the stream.

src/elevenlabs/speech_engine/session.py
```diff
@@ -0,0 +1,521 @@
...
+                    msg = json.loads(raw)
+                except (ValueError, TypeError, UnicodeDecodeError) as e:
+                    await self._emit("error", e)
+                    continue
+
+                await self._handle_message(msg)
```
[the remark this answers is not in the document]
PaulAsjes: Fixed — added a `isinstance(msg, dict)` check after `json.loads()`. Non-dict JSON values now emit an error event and continue the receive loop instead of crashing the session.

src/elevenlabs/speech_engine/session.py
```diff
@@ -0,0 +1,528 @@
...
+            return
+
+        if isinstance(response, str):
+            self._log(
+                'sending string response: "%s", event_id=%s',
+                response,
+                self._current_event_id,
+            )
+            await self._send_agent_response(response, False)
+            await self._send_agent_response("", True)
```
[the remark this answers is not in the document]
PaulAsjes: Fixed — `send_response` now captures `event_id` upfront and passes it explicitly to both `_send_agent_response` calls, matching how `_stream_response` already handled it.

PR #797, opened 2026-05-30
Allow passing `headers, follow_redirects` from `ElevenLabs`, `AsyncElevenLabs` to parent/base class

pchng:
# Summary
- The classes `ElevenLabs` and `AsyncElevenLabs` extend base classes that support the `headers` and `follow_redirects` as constructor parameters, but don't expose these on their own constructors, making it impossible to actually use these.
- Expose these in the child classes and pass onto parent constructors.
- Change should be backwards compatible because parameters are added as optional with default values identical to their parent classes.

---

> [!NOTE]
> **Low Risk**
> Optional constructor kwargs with parent-aligned defaults; behavior change only for callers who previously could not pass these options.
> 
> **Overview**
> **`ElevenLabs` and `AsyncElevenLabs` now accept optional `headers` and `follow_redirects` and pass them to their base classes**, fixing `TypeError` when callers used kwargs the base already supported but the public subclasses did not forward.
> 
> Constructor docs are updated for both sync and async clients. Defaults match the parent (`headers=None`, `follow_redirects=True`), so existing call sites stay unchanged.
> 
> New tests in `test_client_headers.py` cover forwarding to the client wrapper, merged request headers, httpx redirect behavior, and keyword-only signatures for both client types.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e5d34f2cb3e359aadd611374507426a642f17294. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/elevenlabs/client.py
+++ b/src/elevenlabs/client.py
@@ -33,7 +33,11 @@ class ElevenLabs(BaseElevenLabs):
 
         - api_key: typing.Optional[str].
 
+        - headers: typing.Optional[typing.Dict[str, str]]. Additional headers to send with every request.
+
         - timeout: typing.Optional[float]. The timeout to be used, in seconds, for requests by default the timeout is 240 seconds.
+
+        - follow_redirects: typing.Optional[bool]. Whether the default httpx client should follow redirects. Defaults to True. Ignored when a custom `httpx_client` is supplied.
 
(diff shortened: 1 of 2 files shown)


PaulAsjes (approved):
Thanks!

PR #816, opened 2026-07-21
Add uv lockfile for reproducible uv environment setup

sterl27:
## Why
The repository did not have a `uv.lock`, so uv-based environment setup could resolve slightly different dependency versions across machines.

## What changed
This PR adds a generated `uv.lock` based on the existing project dependency configuration so uv installs are reproducible.

## Notes
No runtime source code was changed. This is dependency lock metadata only.

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/bugbot) is generating a summary for commit 39098c394384c0c6dfc01307cf504d5aacbc9e10. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -1,0 +1,394 @@
+version = 1
+revision = 3
+requires-python = ">=3.12"
+
+[[package]]
+name = "annotated-types"
+version = "0.7.0"
+source = { registry = "https://pypi.org/simple" }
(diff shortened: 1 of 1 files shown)


PaulAsjes:
Thanks, but I don't think we need this. This is a library: pip/uv installing elevenlabs from PyPI resolves from the published metadata and never sees any lockfile in the repo. Lockfiles here only pin the dev/CI environment, which poetry.lock already does, and CI (poetry install) consumes it. Adding uv.lock alongside creates two sources of truth for the same thing.

PR #818, opened 2026-07-23
fix: avoid blocking AsyncConversation startup

### LINKED ISSUE #811: AsyncConversation and async client
### Description

The AsyncConversation class inherits from BaseConversation, where get_signed_url is a synchronous method. Calling it from the async implementation may block the event loop

### Code example

https://github.com/elevenlabs/elevenlabs-python/blob/ac134909dc5da91405f4b0c0633cd058d2ef14c8/src/elevenlabs/conversational_ai/conversation.py#L922

### Additional context

_No response_

ogarciarevett:
## Summary

`AsyncConversation.start_session()` currently performs the authenticated signed-URL request synchronously on the event-loop thread. Run that existing synchronous helper in the loop executor so authenticated startup does not pause unrelated async work.

Closes #811.

## Tests

- Added a regression test that verifies signed-URL resolution runs outside the event-loop thread.
- `python -m pytest tests/test_async_convai.py -q` on Python 3.8.20 — 10 passed
- `python tests/verify_types.py` — 1,570 type files passed
- `mypy .` — 2,246 source files passed
- `ruff check tests/test_async_convai.py`

### DIFF
--- a/src/elevenlabs/conversational_ai/conversation.py
+++ b/src/elevenlabs/conversational_ai/conversation.py
@@ -1049,7 +1049,10 @@ class AsyncConversation(BaseConversation):
 
         Will run in background task until `end_session` is called.
         """
-        ws_url = self._get_signed_url() if self.requires_auth else self._get_wss_url()
+        if self.requires_auth:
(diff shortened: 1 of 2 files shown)


PaulAsjes (approved):
Thanks!

PR #823, opened 2026-07-27
fix(stt): send convert form fields unencoded (#819)

### LINKED ISSUE #819: `keyterms` parameter causes `invalid_keyword_length` error in v2.59.0 due to `json.dumps()` serialization regression
### Description

When calling `speech_to_text.convert()` with a keyterms list where every term is well under 50 characters, the API returns a 400 invalid_keyword_length error. This is a regression introduced in v2.59.0.

The cause is in elevenlabs/speech_to_text/raw_client.py — the keyterms list is now wrapped in json.dumps() before being placed in the multipart form body:

`"keyterms": json.dumps(jsonable_encoder(keyterms)),`

This sends a single form field whose value is the full JSON array string (e.g. '["Example1", "Example2", "Example3"]'), rather than repeated individual fields. The API then validates that entire string against the 50-character limit and rejects it.

Steps to reproduce

```
from elevenlabs import ElevenLabs

client = ElevenLabs(api_key="...")
with open("audio.mp3", "rb") as f:
    client.speech_to_text.convert(
        file=f,
        model_id="scribe_v2",
        keyterms=["hello", "world"],  # both well under 50 characters
    )
```

Expected behavior

Each keyterm is sent as a separate repeated form field in the multipart body, as it was in v2.56.0:

```
keyterms=hello
keyterms=world
```

Actual behavior

The entire list is serialized as a JSON string and sent as a single field:

```
keyterms=["hello", "world"]
```

The API returns:

```
{
  "type": "validation_error",
  "code": "invalid_parameters",
  "message": "All keywords must be less than 50 characters.",
  "status": "invalid_keyword_length",
  "param": "keywords"
}
```

Workaround: pin to elevenlabs==2.56.0.

### Code example

_No response_

### Additional context

_No response_

kraenhansen:
Fixes #819.

## Problem

The `SDK regeneration (#817)` regeneration changed four multipart fields of `POST /v1/speech-to-text` from being passed through as-is to being wrapped in `json.dumps(jsonable_encoder(...))`:

```python
"webhook_metadata": json.dumps(jsonable_encoder(webhook_metadata)),
"entity_detection": json.dumps(jsonable_encoder(entity_detection)),
"entity_redaction": json.dumps(jsonable_encoder(entity_redaction)),
"keyterms": json.dumps(jsonable_encoder(keyterms)),
```

That has three consequences on the wire, all of which are visible when dumping the multipart body of `client.speech_to_text.convert(file=..., model_id="scribe_v2", keyterms=["hello", "world"])` on `main`:

```
Content-Disposition: form-data; name="webhook_metadata"

null
Content-Disposition: form-data; name="entity_detection"

null
Content-Disposition: form-data; name="entity_redaction"

null
Content-Disposition: form-data; name="keyterms"

["hello", "world"]
```

1. A list is sent as one field containing the whole JSON array instead of one field per item, so the API validates `["hello", "world"]` as a single keyterm and returns `invalid_keyword_length` — the reported bug.
2. Omitted fields are sent as the literal string `null`, because `jsonable_encoder` maps the `OMIT` sentinel to `None` and `json.dumps(None)` is `"null"`, which is neither `OMIT` nor `None` and therefore survives the request body filtering.
3. A plain string such as `entity_detection="pii"` is sent quoted, as `"pii"`.

## Fix

The generated files can't be edited, so the decoding happens in the hand-maintained `speech_to_text_custom.py`. Both speech-to-text clients now point their raw client at a delegating client wrapper whose HTTP client decodes the affected fields just before the request is sent:

- a JSON array becomes a Python list, which httpx expands into repeated form fields
- a JSON string becomes the plain string
- `null` causes the field to be dropped
- a JSON object (`webhook_metadata`) stays JSON encoded, which is what the API expects

Everything else on the client wrapper and HTTP client is delegated untouched, and only the speech-to-text raw client is affected — the shared client wrapper instance is not mutated, so no other resource sees the interception.

With the fix, the same call sends:

```
Content-Disposition: form-data; name="keyterms"

hello
Content-Disposition: form-data; name="keyterms"

world
```

## Tests

`tests/test_stt_multipart.py` captures the multipart body through an `httpx.MockTransport` and asserts the encoding of each affected field for the sync, async and `with_raw_response` paths, plus a case pinning that unrelated fields are unchanged. Nine of the ten tests fail on `main` and all pass with the fix.

## Relationship to the upstream generator fix

fern-api/fern#17226 fixes this in the generator, but it does not make this wrapper redundant on its own:

- It only covers body properties typed as a list/set of primitives, so it fixes `keyterms` (`Optional[List[str]]`) but not `entity_detection` and `entity_redaction` (`Union[str, List[str]]`) or `webhook_metadata` (`Union[str, Dict[str, Any]]`), which are unions rather than lists and keep the `json.dumps` path, so a plain string stays quoted.
- The separate `"null"` fix (the `is not OMIT else OMIT` guard) shipped in generator 5.0.4, and `.fern/metadata.json` pins `fernapi/fern-python-sdk` to `4.64.1` — the last 4.x release, published two days before 5.0.4. Picking up either fix requires bumping the generator to 5.x.

Decoding only touches string values, so each field stops being rewritten as soon as the generator stops encoding it: after the generator upgrade `keyterms` arrives as a list and omitted fields arrive as `OMIT`, neither of which is a string. The wrapper can be deleted once all four fields are generated correctly.

## Not covered

`dubbing.project.create(keyterms=...)`, `music.compose(tags=...)`, `voices.ivc.create(labels=...)`, `studio.projects.create(genres=...)` and `audio_native.create(...)` got the same `json.dumps` treatment in the same regeneration. Those resources have no hand-maintained wrapper to hook into, so they are left for the generator upgrade.

<div><a href="https://cursor.com/agents/bc-99983d8a-f0f3-4d6b-a98c-2900d639fa07"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-99983d8a-f0f3-4d6b-a98c-2900d639fa07"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


kraenhansen:
Closing and waiting for an upstream fix by the Fern team 🤞

PR #839, opened 2026-08-10
fix: preserve ConvAI audio replies after interruptions

### LINKED ISSUE #829: ConvAI: reply to send_multimodal_message is silently dropped — its audio events share the triggering interruption's event_id and fail the <= gate
### Description

**Environment:** `elevenlabs` 2.59.0 (drop logic unchanged on current `main` @ a1a40fc), Python 3.12, ConvAI `Conversation` over WebSocket with an `AudioInterface` and `client_tools`.

**What we were doing:** A voice agent (running on a robot) has a client tool that captures a camera frame, uploads it with `client.conversational_ai.conversations.files.create(...)`, and hands it to the live conversation via `conversation.send_multimodal_message(text=..., file_id=...)` from inside the tool handler.

**Actual behavior:** The agent's spoken reply to that multimodal message is completely silent, with no error anywhere. The `callback_agent_response` text callback fires with the correct answer, but `audio_interface.output()` is never called. Client-side instrumentation shows why:

1. `send_multimodal_message` lands → the server fires an `interruption` event ~60 ms later.
2. The reply's `audio` events then arrive carrying an `event_id` **equal to** that interruption's `event_id`.
3. `_handle_message_core` drops audio events with `event_id <= self._last_interrupt_id`, so every chunk of the reply is discarded:

https://github.com/elevenlabs/elevenlabs-python/blob/a1a40fcaf1dc4d8b7bc8aab55c32b0f4a04502d2/src/elevenlabs/conversational_ai/conversation.py#L530

(same in the async path at L600; the gate is raised by the interruption handler at L572–575.)

Captured sequence (timestamps real, ids from one session):

```
15:28:39.712  [client tool] frame uploaded; send_multimodal_message(...)
15:28:39.772  interruption event: {'event_id': 87}   <- fired by the server for our own message
15:28:41.441  audio event id=87  gate=87  -> dropped by the <= check
15:28:41.563  audio event id=87  gate=87  -> dropped
15:28:41.745  agent_response text arrives normally (text has no event-id gate)
...           every audio chunk of the reply carries id=87; none is delivered
```

This reproduced 3/3 times within the session (interruption/audio id pairs 87/87, 140/140, 210/210). It is **race-dependent**, which makes it look intermittent: when concurrent real user speech advances the event counter first, the reply's audio ids exceed the gate and play fine (also captured: `interruption id=22` → reply audio `id=27` → audible).

**Expected behavior:** The spoken reply to a `send_multimodal_message` should be audible — its audio events should not be filtered out by the interruption that the same message triggered. At minimum, dropped audio events would deserve a debug log; today the failure is entirely silent.

**Steps to reproduce:**
1. Start a `Conversation` with any `audio_interface` against an agent.
2. Mid-session, while no user audio is active, upload a file and call `conversation.send_multimodal_message(text="What is in this image?", file_id=...)` (in our case from a client-tool handler, but that isn't essential).
3. Observe the `interruption` event, then the reply's `audio` events sharing its `event_id` — `output()` is never called for them while the text callbacks fire normally.

### Code example

Minimal instrumentation that makes the drop visible (wraps the message handler; API key not involved):

```python
def install_audio_gate_logger(conversation):
    original = conversation._handle_message

    def wrapped(message, ws):
        if message.get("type") == "audio":
            event_id = int(message["audio_event"]["event_id"])
            gate = conversation._last_interrupt_id
            if event_id <= gate:
                print(f"audio event id={event_id} gate={gate} -> WILL BE DROPPED")
        elif message.get("type") == "interruption":
            print(f"interruption: {message['interruption_event']}")
        return original(message, ws)

    conversation._handle_message = wrapped

### Additional context

- Two candidate fixes:
  - **Client-side:** use a strict `<` comparison. In all our captures, audio that was genuinely in flight when an interruption fired always had an id strictly below the interruption's id (the interruption is a later event in the stream), and all chunks of one response share a single event_id — so `id == gate` only ever described the *new* reply that the interruption's own trigger produced.
  - **Server-side:** number a reply's audio events after the interruption produced by its triggering message.
- Workaround we currently ship (delivers a fresh reply numbered at the gate, while keeping genuine barge-in drops — stale audio has ids ≤ the pre-interruption maximum, which the shim tracks):

```python
def install_gate_fix(conversation):
    original = conversation._handle_message
    state = {"max_audio_id": 0, "max_audio_id_at_interrupt": 0}

    def wrapped(message, ws):
        mtype = message.get("type")
        if mtype == "audio":
            event_id = int(message["audio_event"]["event_id"])
            gate = int(conversation._last_interrupt_id)
            if event_id == gate and event_id > state["max_audio_id_at_interrupt"]:
                conversation._last_interrupt_id = event_id - 1  # deliver this reply
            state["max_audio_id"] = max(state["max_audio_id"], event_id)
        elif mtype == "interruption":
            state["max_audio_id_at_interrupt"] = state["max_audio_id"]
        return original(message, ws)

    conversation._handle_message = wrapped
```

- Happy to provide fuller logs or run diagnostics against a test build.

uuzzrm:
## Summary

Fixes #829.

ConvAI can send the interruption event and the audio events for the resulting reply with the same `event_id`. The sync and async handlers currently discard those audio events because they reject IDs less than or equal to the last interruption. This changes both checks to reject only older IDs, so equal-ID reply audio is delivered while stale audio remains filtered.

## Reference

This matches the official JS client (@elevenlabs/client), which plays audio when `lastInterruptTimestamp <= audio_event.event_id` and starts `lastInterruptTimestamp` at 0 - in other words, drop strictly older audio only. The Python gate now applies the same rule on `_last_interrupt_id`, including the edge case where an audio event carries id 0 before any interruption.

## Changes

- Update the sync and async audio-event guards in the hand-maintained ConvAI wrapper.
- Add regression coverage for equal-ID reply audio and older audio events in both paths.

## Testing

- `PYTHONPATH=src uv run --with pytest --with pytest-asyncio --with websockets pytest tests/test_convai.py tests/test_async_convai.py tests/test_convai_audio_events.py -q`
- `PYTHONPATH=src uv run --no-project python3 -m compileall -q src/elevenlabs/conversational_ai/conversation.py tests/test_convai_audio_events.py`
- Offline suite: `171 passed` after excluding the existing API-backed integration tests, which require `ELEVENLABS_API_KEY`.
- `uv run --with ruff ruff check tests/test_convai_audio_events.py`
- `git diff --check`

The repository README notes that direct SDK changes are proof-of-concept contributions because most of the SDK is generated by Fern; this PR is limited to files listed in `.fernignore`.

## Follow-up (review feedback)

- Added the strictly-newer audio case (88 after interrupt 87) for both sync and async paths. The previous set only pinned `=` and `<`, so a rule that had regressed to equality-only would have slipped through.
- Renamed the tests around the drop rule itself so the parity with the JS client is legible to the next reader.
- Focused run: `pytest tests/test_convai_audio_events.py -q` - 6 passed.

---

> [!NOTE]
> **Low Risk**
> Narrow behavioral fix in ConvAI message handling with targeted unit tests; no auth, data, or API surface changes.
> 
> **Overview**
> Fixes dropped ConvAI reply audio when the server sends **interruption** and **audio** events sharing the same `event_id`.
> 
> In `conversation.py`, sync and async audio handlers now skip only events with `event_id` **strictly less than** `_last_interrupt_id` (was `<=`), so post-interruption reply audio at the same ID is played while pre-interruption audio is still filtered.
> 
> Adds `tests/test_convai_audio_events.py` with sync/async regression tests for equal-ID, newer, and older audio relative to an interruption.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 318ad071bac196fc816c8f72c723e771b0f50288. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/elevenlabs/conversational_ai/conversation.py
+++ b/src/elevenlabs/conversational_ai/conversation.py
@@ -527,7 +527,7 @@ class BaseConversation:
 
         elif message["type"] == "audio":
             event = message["audio_event"]
-            if int(event["event_id"]) <= self._last_interrupt_id:
+            if int(event["event_id"]) < self._last_interrupt_id:
                 return
             audio = base64.b64decode(event["audio_base_64"])
             message_handler.handle_audio_output(audio)
@@ -597,7 +597,7 @@ class BaseConversation:
 
         elif message["type"] == "audio":
             event = message["audio_event"]
-            if int(event["event_id"]) <= self._last_interrupt_id:
+            if int(event["event_id"]) < self._last_interrupt_id:
                 return
             audio = base64.b64decode(event["audio_base_64"])
             await message_handler.handle_audio_output(audio)
--- a/tests/test_convai_audio_events.py
+++ b/tests/test_convai_audio_events.py
@@ -1,0 +1,76 @@
+import base64
+from unittest.mock import AsyncMock, MagicMock
+
+import pytest
+
+from elevenlabs.conversational_ai.conversation import BaseConversation
+
+
+def _conversation_with_interrupt(event_id: int) -> BaseConversation:
+    conversation = BaseConversation.__new__(BaseConversation)
+    conversation._last_interrupt_id = event_id
+    return conversation
+
+
+def _audio_message(event_id: int) -> dict:
+    return {
+        "type": "audio",
+        "audio_event": {
+            "event_id": str(event_id),
+            "audio_base_64": base64.b64encode(b"reply audio").decode("ascii"),
+        },
+    }
+
+
+def _interruption_message(event_id: int) -> dict:
+    return {
+        "type": "interruption",
+        "interruption_event": {"event_id": str(event_id)},
+    }
+
+
+def test_sync_conversation_keeps_audio_with_interruption_event_id():
+    conversation = _conversation_with_interrupt(0)
+    message_handler = MagicMock()
+    message_handler.callback_audio_alignment = None
+
+    conversation._handle_message_core(_interruption_message(87), message_handler)
+    conversation._handle_message_core(_audio_message(87), message_handler)
+
+    message_handler.handle_audio_output.assert_called_once_with(b"reply audio")
+
+
+def test_sync_conversation_discards_audio_older_than_interruption():
+    conversation = _conversation_with_interrupt(87)
+    message_handler = MagicMock()
+    message_handler.callback_audio_alignment = None
+
+    conversation._handle_message_core(_audio_message(86), message_handler)
+
+    message_handler.handle_audio_output.assert_not_called()
+
+
+@pytest.mark.asyncio
+async def test_async_conversation_keeps_audio_with_interruption_event_id():
+    conversation = _conversation_with_interrupt(0)
+    message_handler = MagicMock()
+    message_handler.callback_audio_alignment = None
+    message_handler.handle_audio_output = AsyncMock()
+    message_handler.handle_interruption = AsyncMock()
+
+    await conversation._handle_message_core_async(_interruption_message(87), message_handler)
+    await conversation._handle_message_core_async(_audio_message(87), message_handler)
+
+    message_handler.handle_audio_output.assert_awaited_once_with(b"reply audio")
+
+
+@pytest.mark.asyncio
+async def test_async_conversation_discards_audio_older_than_interruption():
+    conversation = _conversation_with_interrupt(87)
+    message_handler = MagicMock()
+    message_handler.callback_audio_alignment = None
+    message_handler.handle_audio_output = AsyncMock()
+
+    await conversation._handle_message_core_async(_audio_message(86), message_handler)
+
+    message_handler.handle_audio_output.assert_not_awaited()
--- a/tests/test_convai_audio_events.py
+++ b/tests/test_convai_audio_events.py
@@ -29,7 +29,7 @@ def _interruption_message(event_id: int) -> dict:
     }
 
 
-def test_sync_conversation_keeps_audio_with_interruption_event_id():
+def test_sync_conversation_keeps_audio_equal_to_interruption():
     conversation = _conversation_with_interrupt(0)
     message_handler = MagicMock()
     message_handler.callback_audio_alignment = None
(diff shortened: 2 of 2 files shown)


maksym-t-didww:
Checked this against the reference implementation, and it backs the change: after this PR the Python gate matches the official JS client exactly.

`@elevenlabs/client` applies the rule on play rather than on drop — `packages/client/src/VoiceConversation.ts`:

```ts
if (this.lastInterruptTimestamp <= event.audio_event.event_id) {
  // ... plays
}
```

with `lastInterruptTimestamp` initialised to `0` and assigned in `BaseConversation.handleInterruption` from `interruption_event.event_id` — the same lifecycle as `_last_interrupt_id` here.

So the reference rule is *drop strictly older*: `audio_id < lastInterrupt`. The `<=` here drops one extra state, equality, which is exactly the silence reported in #829. This PR closes that gap.

It also removes a second, quieter divergence: `_last_interrupt_id` starts at `0`, so today an `audio` event with `event_id` 0 is dropped before any interruption has occurred, while the JS client plays it (`0 <= 0`). Evidently `event_id` 0 does not occur in practice, since nobody has reported it, but the same one-character change makes the two agree there too.

Two cheap suggestions for the tests:

1. There is no case for audio strictly newer than the interrupt (`88` after `87`) — the ordinary post-interruption reply. The current pair pins `=` and `<` but not `>`, so a rule that had accidentally become `==` would still pass both.
2. Consider naming the tests after the rule rather than the symptom, e.g. `test_sync_conversation_drops_only_audio_older_than_interruption`. It makes the parity with the JS client legible to the next reader.

Might be worth putting the JS citation in the PR description as well — the justification currently rests on the captured session in #829, and this turns it from an empirical claim into a protocol one.

uuzzrm:
Appreciate the detailed check against the JS client. I added the strictly-newer case (88 after 87) for both paths and renamed the tests around the rule itself, and put the @elevenlabs/client reference in the PR description so the justification stands on the protocol rather than the captured session alone.

Focused run: pytest tests/test_convai_audio_events.py -q - 6 passed.

PR #858, opened 2026-09-02
Stop the OMIT sentinel from reaching the realtime TTS connection

Kayvan-Zahiri:
`convert_realtime` raises `AttributeError: 'ellipsis' object has no attribute 'dict'` at its documented default, before it sends any text. `model_id` and `voice_settings` both default to `OMIT`, which is `Ellipsis`. Line 115 is `voice_settings.dict() if voice_settings else None`, and `Ellipsis` is truthy, so `.dict()` is called on the sentinel.

The same sentinel reaches the URL. `model_id` goes straight into `build_ws_url`, where `urlencode` stringifies it, so a caller who passes `voice_settings` but omits `model_id` connects to `stream-input?model_id=Ellipsis&output_format=mp3_44100_128` instead of letting the server apply its default. This predates the `build_ws_url` refactor in #780; the old f-string produced the same query string.

`realtime/scribe.py` and `conversational_ai/conversation.py` both build their params conditionally. This does the same, reusing the `remove_none_from_dict` already imported in this file.

Four of the six added tests fail on `main`. The suite goes from 189 to 195 passed, with the same 16 live-API failures on either side, and `mypy` is clean. Both files touched are listed in `.fernignore`. I left the pre-existing `ruff` import-order warning in this file alone rather than bury a six line fix.

### DIFF
--- a/src/elevenlabs/realtime_tts.py
+++ b/src/elevenlabs/realtime_tts.py
@@ -92,11 +92,17 @@ class RealtimeTextToSpeechClient(TextToSpeechClient):
             ),
         )
         """
+        # OMIT is a sentinel meaning "not supplied", so it must never reach the wire
+        if model_id is OMIT:
+            model_id = None
+        if voice_settings is OMIT:
+            voice_settings = None
+
         with connect(
             build_ws_url(
                 self._ws_base_url,
                 ["v1", "text-to-speech", voice_id, "stream-input"],
-                {"model_id": model_id, "output_format": output_format},
+                remove_none_from_dict({"model_id": model_id, "output_format": output_format}),
             ),
             additional_headers=jsonable_encoder(
                 remove_none_from_dict(
--- a/tests/test_realtime_tts_params.py
+++ b/tests/test_realtime_tts_params.py
@@ -1,0 +1,67 @@
+"""convert_realtime must not leak the OMIT sentinel into the URL or the first frame."""
+
+import base64
+import json
+from unittest.mock import MagicMock, patch
+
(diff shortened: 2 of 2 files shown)


PaulAsjes:
Thanks for the contribution, but the file you've edited is autogenerated so your changes will be overwritten the next time we generate the SDKs. 

Instead this needs to be fixed upstream in the generator, so I'm closing this for now.

Kayvan-Zahiri:
Thanks for taking a look, and no problem either way. I think this one may be a false alarm
though, so flagging it in case it's worth a second look.

`src/elevenlabs/realtime_tts.py` is listed in `.fernignore`, under "Ignore manually created
SDK wrappers", and `tests/` is ignored too, so both files in this PR are excluded from
regeneration. `AGENTS.md` says "Only files listed in `.fernignore` are hand-maintained and
safe to edit", which is how I picked it.

The file does still carry the "auto-generated by Fern" header, which is what I suspect
prompted the close. That header looks stale rather than accurate:

- The last `fern-api[bot]` commit to touch it was July 21, 2025.
- PR #780 (April 29, 2026) hand-refactored it to use `urllib.parse` for WebSocket URL
  construction, and `import urllib.parse` is still there on main today, after the
  May 20, 2026 SDK regeneration in #792.

So a human edit to this file has already survived a regeneration cycle.

Happy to be wrong here. If the OMIT handling really does originate in the generator I'll
leave it alone, and if it would help I can open an issue describing the bug instead so it
is tracked somewhere useful.

PaulAsjes:
You're absolutely right, sorry about that. Will look at this again in more detail.

PaulAsjes (approved):
Thanks!

PR #862, opened 2026-09-06
fix: expose WebSocket close metadata in RealtimeEvents.CLOSE handler

Shreyaaaash:
- Pass close_code and close_reason to CLOSE event handlers
- Update docstring to document the new callback signature for CLOSE events
- Fixes issue where CLOSE handlers received zero arguments, making it impossible to distinguish normal closes from errors


Shreyaaaash:
### Addressing Cursor Bot's Version Compatibility Concern

Cursor Bot correctly pointed out that the initial implementation relied on `websocket.close_code` and `websocket.close_reason`, which were introduced in `websockets` 14.1. Since the project supports `websockets >=13.0`, those attributes cannot be relied upon.

The implementation has been updated to address this compatibility issue.

Instead of accessing close metadata directly from the WebSocket object, the updated implementation catches the `ConnectionClosed` exception and extracts the close information from the exception itself:

```python
except ConnectionClosed as e:
    # Extract close metadata from the received close frame.
    if hasattr(e, "rcvd") and e.rcvd is not None:
        self._close_code = e.rcvd.code
        self._close_reason = e.rcvd.reason
    elif hasattr(e, "sent") and e.sent is not None:
        # Fall back to the sent close frame when no received frame is available.
        self._close_code = e.sent.code
        self._close_reason = e.sent.reason

    # Emit an error for abnormal WebSocket closures.
    if isinstance(e, ConnectionClosedError) and e.rcvd is not None:
        self._emit(
            RealtimeEvents.ERROR,
            {
                "error": (
                    f"WebSocket connection closed with error: "
                    f"{e.rcvd.code} {e.rcvd.reason}"
                )
            },
        )
```

The close code and reason are stored in instance variables (`_close_code` and `_close_reason`) so they remain available to the `finally` block when the `CLOSE` event is emitted.

### Why this is compatible

The `rcvd` and `sent` attributes on `ConnectionClosed` exceptions are available in the supported `websockets` versions, including the project's minimum version of 13.0. Therefore, the implementation no longer depends on the `websocket.close_code` and `websocket.close_reason` attributes introduced in 14.1.

A fallback to the WebSocket attributes can still be retained as a defensive measure, but they are no longer part of the primary code path.

### Result

`CLOSE` event handlers now receive:

```text
close_code: int | None
close_reason: str | None
```

This allows applications to distinguish between normal and abnormal WebSocket closures:

* `close_code == 1000` — normal provider-initiated closure.
* `close_code != 1000` — abnormal closure that may require recovery or error handling.
* `close_reason` — the provider's close message, when available.

This approach preserves the required close metadata while remaining compatible with the project's `websockets >=13.0` dependency constraint.

PR #869, opened 2026-09-16
fix(play): raise when ffplay exits with an error

### LINKED ISSUE #868: [play.py] `Play` silently failing if ffplay is installed but broken
### Description

The quick-start example using 'play' was running but there was no output. `is_installed("ffplay")` was returning true. It turns out ffplay was installed but but broken on my machine and  `play` function did not catch the failure. 

https://github.com/elevenlabs/elevenlabs-python/blob/bc7f996e5a0a92a5b1feb09503a89db0dd70dc83/src/elevenlabs/play.py#L38-L46

the `returncode` and `err` should be used to catch those type of errors
```
if proc.returncode != 0:
            message = (
                f"ffplay failed (exit {proc.returncode}): {err.decode(errors='replace')}"
            )
            raise RuntimeError(message)
```

**Steps to Reproduce**
1. Simulate a broken ffplay binary by running this in your terminal:  
   ```printf '#!/bin/sh\nexit 1\n' > /tmp/ffplay && chmod +x /tmp/ffplay```
3. Call `play` using that binary by running the following on the terminal: 
```PATH="/tmp:$PATH" uv run python -c "from elevenlabs.play import play; play(b'x')"```

**Expected Behavior** 
`play()` should throw an error.

** Actual Behavior ** 
`play()` silently fails and there is no error and no output

### Code example
```
import os

from dotenv import load_dotenv
from elevenlabs.client import ElevenLabs
from elevenlabs.play import is_installed, play

load_dotenv()
elevenlabs = ElevenLabs(
    api_key=os.getenv("ELEVENLABS_API_KEY"),
)
audio = elevenlabs.text_to_speech.convert(
    text="The first move is what sets everything in motion.",
    voice_id="JBFqnCBsd6RMkjVDRZzb",  # "George" - browse voices at elevenlabs.io/app/voice-library
    model_id="eleven_v3",
    output_format="mp3_44100_128",
)
print(is_installed("ffplay"))
play(audio)
```

**Running ffplay in terminal:**

`ffplay`         
```                                
dyld[38433]: Library not loaded: /opt/homebrew/opt/jpeg-xl/lib/libjxl.0.10.dylib
  Referenced from: <D983C040-DBB4-3CC6-BDF9-36D2867B927A> /opt/homebrew/Cellar/ffmpeg/7.0/bin/ffplay
  Reason: tried: '/opt/homebrew/opt/jpeg-xl/lib/libjxl.0.10.dylib' (no such file), '/System/Volumes/Preboot/Cryptexes/OS/opt/homebrew/opt/jpeg-xl/lib/libjxl.0.10.dylib' (no such file), '/opt/homebrew/opt/jpeg-xl/lib/libjxl.0.10.dylib' (no such file), '/usr/local/lib/libjxl.0.10.dylib' (no such file), '/usr/lib/libjxl.0.10.dylib' (no such file, not in dyld cache), '/opt/homebrew/Cellar/jpeg-xl/0.11.1/lib/libjxl.0.10.dylib' (no such file), '/System/Volumes/Preboot/Cryptexes/OS/opt/homebrew/Cellar/jpeg-xl/0.11.1/lib/libjxl.0.10.dylib' (no such file), '/opt/homebrew/Cellar/jpeg-xl/0.11.1/lib/libjxl.0.10.dylib' (no such file), '/usr/local/lib/libjxl.0.10.dylib' (no such file), '/usr/lib/libjxl.0.10.dylib' (no such file, not in dyld cache)
zsh: abort      ffplay
```
### Additional context

Running `brew upgrade ffmpeg` fixed the ffplay issue for me, but raising this issue so that the API wouldnt silently fail.

_No response_

### PR DESCRIPTION
**Pull request by dchaudhari7177**
Fixes #868.

`play()` pipes the audio to `ffplay` and then called `proc.poll()` without looking at the result. The `is_installed("ffplay")` check only proves the binary is on `PATH`. If ffplay is there but can't run (a build missing a shared library, no usable audio device), `play()` returned normally and nothing was heard, which is what the issue reports.

After `communicate()`, a non-zero `returncode` now raises `ValueError` (the error type `play()` already uses for a missing ffplay or a missing optional dependency). The message gives the exit status and the tail of ffplay's stderr, so the cause is visible:

```text
ValueError: ffplay exited with status 1 and did not play the audio. Its error output was:
error while loading shared libraries: libavdevice.so.61
```

`play.py` is listed in `.fernignore`, so this edits the hand-maintained wrapper, not generated code.

### Tests

New `tests/test_play.py`. `play.py` had no tests. `subprocess.Popen` is replaced with a real Python subprocess that reads the piped audio from stdin, writes stderr and exits with a chosen code, so the pipe handling runs for real:

- ffplay exits 0: `play()` returns and `ffplay` is the command it launched
- exit 1 with stderr: raises, and the message carries the stderr text
- exit 2 with no stderr: raises with the status only
- an iterator of chunks is joined and piped

```text
$ pytest tests/test_play.py
this branch:                  4 passed
play.py restored from main:   2 failed, 2 passed   (both failure cases: DID NOT RAISE)
$ ruff check src/elevenlabs/play.py tests/test_play.py   (ruff 0.11.5, as pinned)
All checks passed!
```


PaulAsjes (approved):
Thanks!

PR #871, opened 2026-09-21
Add transcript_edit to the realtime speech-to-text wrapper

JacekDuszenko:
Adds the `transcript_edit` option to `client.speech_to_text.realtime.connect()` (both `RealtimeAudioOptions` and `RealtimeUrlOptions`) and the `RealtimeEvents.EDITED_TRANSCRIPT` event, following the pattern of #830 (entity detection).

- `transcript_edit: str`: natural-language instruction applied to each committed transcript (max 2000 characters), sent as the `transcript_edit` query param. Cannot be combined with `entity_detection`; the server rejects the combination with `invalid_request`.
- New server message `edited_transcript` → `RealtimeEvents.EDITED_TRANSCRIPT`, carrying `text` (the committed text the instruction was applied to) and `edited_text` (identical to `text` if no edits were made). Edits run asynchronously and may arrive out of commit order, so `text` is the field to correlate on. Dispatch is by enum value, so adding the member is what routes the message.
- Tests: the exhaustive URL-builder and connect() tests cover `transcript_edit`; dispatch tests cover the new event.
- Version bumped to 2.69.0 (`pyproject.toml`, `client_wrapper` headers), as #830 did. Happy to drop that if releases are cut differently now.

Upstream: the parameter and event ship in elevenlabs/xi#49618 (websocket route) and elevenlabs/xi#51266 (public AsyncAPI spec); docs in elevenlabs/elevenlabs-dx#2778; JS counterpart in elevenlabs/elevenlabs-js (same branch name); client SDK in elevenlabs/packages#1023. Pricing: 30% premium on the base transcription cost, billed for at least 10 seconds of audio per committed transcript.

Verified locally: `ruff check` clean, `mypy src/elevenlabs/realtime` clean, `pytest tests/test_stt_realtime.py` 33 passed.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


src/elevenlabs/core/client_wrapper.py
```diff
@@ -27,12 +27,12 @@ def get_headers(self) -> typing.Dict[str, str]:
         import platform
 
         headers: typing.Dict[str, str] = {
-            "User-Agent": "elevenlabs/2.68.0",
+            "User-Agent": "elevenlabs/2.69.0",
```
PaulAsjes: Don't need this, the next release will automate the version.

pyproject.toml
```diff
@@ -4,7 +4,7 @@ dynamic = ["version"]
 
 [tool.poetry]
 name = "elevenlabs"
-version = "2.68.0"
+version = "2.69.0"
```
PaulAsjes: Same, don't need this.

PaulAsjes (approved):

